PR #7, opened 2025-10-25
Standalone home and other changes

### LINKED ISSUE #5: standalone home-manager
@vic do you think this flake should support standalone home-manager configurations? I don't use it myself, but probably there are people out there that use it

HeitorAugustoLN:
Closes #5

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,8 +107,8 @@ For standalone home-manager configurations (without a NixOS/Darwin host):
 # modules/homes.nix -- see <den>/nix/types.nix for schema.
 {
   # Define standalone home configurations for any system.
-  den.homes.x86-64-linux.vic = {};
-  
+  den.homes.x86_64-linux.vic = {};
+
   # Multiple homes can share the same aspect.
   den.homes.aarch64-darwin.vic = {};
 }


nix/types.nix
```diff
@@ -39,11 +42,13 @@ let
   userType = lib.types.submodule (
     { name, ... }:
     {
+      freeformType = lib.types.attrsOf lib.types.anything;
       options = {
         name = strOpt "user configuration name" name;
         userName = strOpt "user account name" name;
         class = strOpt "home management nix class" "homeManager";
         aspect = strOpt "main aspect name" name;
+        homeManager = nullableStrOpt "home-manager input name override";
```
HeitorAugustoLN: This is here for consistency sakes only, it doesn't get used. Also should inputs be defined as the actual input or just the name?

HeitorAugustoLN: will drop this changes for now

nix/aspects-config.nix
```diff
@@ -18,6 +19,16 @@ let
       };
     };
 
+  homeAspect =
+    home:
+    { aspects, ... }:
+    {
+      ${home.aspect} = {
+        includes = [ (aspects.default.home { inherit home; }) ];
+        homeManager = { };
```
vic: let's use `${home.class}` here instead of hardcoded `homeManager`.

nix/types.nix
```diff
@@ -18,6 +18,7 @@ let
     lib.types.submodule (
       { name, config, ... }:
       {
+        freeformType = lib.types.attrsOf lib.types.anything;
```
vic, line 21: this is awesome! can you please update the README to mention that `host`, `user` and `home` objects have freeforms (people can add whatever attributes they need) and they can be used in aspect provider functions (those registered on `flake.aspects.default.{home,user,host}` ?

nix/flakeModule.nix
```diff
@@ -5,7 +5,9 @@ in
 {
   imports = [
     ./os-config.nix
+    ./home-config.nix
     ./aspects-config.nix
   ];
-  options.den = types.denOption;
+  options.den.hosts = types.hostsOption;
```
vic, line 11: nice!, please also update the readme examples that used `den.<system>` to use `den.hosts.<system>`

nix/home-config.nix
```diff
@@ -0,0 +1,31 @@
...
+        inherit pkgs;
+        modules = [ self.modules.${home.class}.${home.aspect} ];
+      }
+    );
+  };
+
+  homeConfigurations =
+    class: builtins.listToAttrs (map homeConfiguration (builtins.filter (x: x.class == class) homes));
+
+  flake.homeConfigurations = homeConfigurations "homeManager";
```
vic, line 27: this is correct, however it looks like our current flake checks on the default template are not verifying the homeConfiguration. I guess because the default flake has no dependency on homeManager itself, can you please add another module inside `_example` that adds `flake-file.inputs.home-manager` dependency ? so we can properly check the example configuration ?

vic (changes requested):
wow ! looks awesome!

I left some changes request, thank you again :)

vic:
Thank you, @HeitorAugustoLN

PR #8, opened 2025-10-26
support different inputs for nixpkgs, darwin. also add systemManager.

vic:
host and home options now allow setting a custom instantiate function to allow overriding the input or the function used to build the OS/home if people **really** need things like specialArgs

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -26,4 +26,7 @@ jobs:
           }
           EOF
           nix run .#write-flake
+          nix flake update den
+          nix run .#write-flake
+          nix flake metadata
           nix flake check


vic:
regarding specialArgs, @HeitorAugustoLN.

vic:
For example some people (like me) that do use homeManager inside a nixos configuration would still like to have a `homeConfiguration.vic@host` so that it can be applied without re-building the whole OS. see: https://github.com/mightyiam/dendritic/discussions/11#discussioncomment-14651266

HeitorAugustoLN:
How does adding specialArgs work?

vic:
I've just documented it at README.

PR #9, opened 2025-10-26
refactor(config): remove deprecated top-level system argument

HeitorAugustoLN:
Removes the deprecated top-level `system` argument from OS configurations.

This change aligns with modern NixOS practice, where the system is defined inside modules (e.g., via `nixpkgs.hostPlatform`). This improves build reproducibility.

If we want to still keep a similar behavior, we could add a module option by default to set: `{ nixpkgs.hostPlatform = lib.mkDefault host.system; }`

Related: NixOS/nixpkgs#177012

### DIFF
--- a/nix/config.nix
+++ b/nix/config.nix
@@ -21,8 +21,10 @@ let
   osConfiguration =
     host:
     host.instantiate {
-      inherit (host) system;
-      modules = [ self.modules.${host.class}.${host.aspect} ];
+      modules = [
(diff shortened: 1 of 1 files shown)


vic:
Nice!

+1 on adding default `nixpkgs.hostPlatform`, perhaps in _example for now so checks can pass.

vic:
Thanks, @HeitorAugustoLN :+1:

PR #14, opened 2025-10-27
Use den.aspects/den.modules instead of flake.aspects/flake.modules

vic:
Using den scope allows cleaner separation of aspects, so we dont polute generic flake.modules.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ Need more batteries? see [vic/denful](https://github.com/vic/denful)
   # $ home-manager switch --flake .#vic
   den.homes.aarch64-darwin.vic = { };
 
-  # That's it! The rest is adding flake.aspects.
+  # That's it! The rest is adding den.aspects.
 }
 ```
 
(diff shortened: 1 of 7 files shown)


HeitorAugustoLN:
When should I use flake.aspects instead of den.aspects? I think I would prefer to keep it all on flake.aspects

vic:
@HeitorAugustoLN,

The reason why den is not using the `flake.aspects` **attribute**:

**`flake.aspects` attribute IS a transposition of `flake.modules`**.

Everything you add to `flake.aspects` will be automatically available at `flake.modules`.
Sometimes this is the thing you want, like if you are writing modules for other people to use outside your flake. Since `flake.modules` is **exposed** as part of your flake output attributes. The flake-aspects project makes it easy do expose such modules from resolved aspects. 

However, for `den`, this is something we don't always want. Because for each host, user, home, `den` will automatically create an aspect for them, (and I have 11 hosts, two users), and dont want those modules to be part of my flake outputs. 

If we used `flake.aspects` attribute the `flake.modules` output would be really crowded.

Because of this, den creates its own aspects scope (see modules/scope.nix), it defines its own `den.aspects` and `den.modules`.

Having the os-configurations read from `den.modules` instead of `flake.modules` also makes the library more flake independent. (I'm experimenting with non-flake dendritic setups -- this is the reason why flake-aspects library itself is also usable outside flakes). 

You can keep using `flake.aspects` if you want, just have a module that does:

```
{ config, ...}:
{ 
  den.aspects = config.flake.aspects;
}
```

they are independent, so you can do it the other way around: `flake.modules = den.modules;` if you wanted to publish all your den modules.

vic:
Also, you can always use `includes` with any aspect independently of where they are defined:

```
den.aspects.my-laptop.includes = [ config.flake.aspects.something ]; 
```

PR #79, opened 2025-11-12
example namespaces and angle-brackets in default template.

vic:
split into several files, more dendritic.


vic:
@xdbr guess what !? Looks like you discovered a bug, haha. Because there was no test exercising the angle brackets syntax. So I just updated our default template in this PR and it uncovered a bug that angle-brackets did not correctly handled namespaces. Fix is already in this commit.

xdbr:
@vic well... i'm happy to help ;) after all, there seems to be value in being an early adopter – plus you get front-row seats ;) looking forward to your fix!

vic:
> @vic well... i'm happy to help ;) after all, there seems to be value in being an early adopter – plus you get front-row seats ;) looking forward to your fix!

already merged, angle-brackets should work now with namespaces. 

I'm not sure about your `"foo.bar"` naming tho, that might cause problems, I'm not sure that <vix/"foo.bar"> is valid nix syntax, I guess it is not.

xdbr:
Sadly, even after pulling the latest update (through `nix flake update`) the fix doesn't (yet) work for me in my [configuration.nix](https://github.com/xdbr/fleet/blob/dev/modules/hosts/barbara.bar/configuration.nix)

Some questions and observations:
- [ ] what exactly is the difference between `imports = [ (inputs.den.namespace "eg" false) ];` and `imports = [ (inputs.den.namespace "eg" true) ];`? I believe the documentation uses `true`?
- [ ] is the full exposition exercised, i.e. the combination of `__findFile`, custom namespace and an `includes` with angle-bracket syntax? 
- [ ] if not, could [igloo.nix](https://github.com/vic/den/blob/f57da63114e2ccc2d465a6dc954192f0eb74d1b3/templates/default/modules/aspects/igloo.nix) potentially be a good place to rewrite into this direction to see if it really works in this kind of setup/situation?
- [ ] also, is there a (technical) possibility to have `__findFile` available automatically? however, this is obviously only a cosmetic improvement, but as far as I can see, apparently it is possible to have a custom namespace automatically in scope, so maybe yes?!

xdbr:
> I'm not sure about your `"foo.bar"` naming tho, that might cause problems, I'm not sure that <vix/"foo.bar"> is valid nix syntax, I guess it is not.

Ah I see – just to mention, I haven't used `<vix/"foo.bar">`-type includes but I can try to see if renaming the aspect helps, just to make sure

vic:
> Sadly, even after pulling the latest update (through `nix flake update`) the fix doesn't (yet) work for me in my [configuration.nix](https://github.com/xdbr/fleet/blob/dev/modules/hosts/barbara.bar/configuration.nix)
> 
> Some questions and observations:
> 
>     * [ ]  what exactly is the difference between `imports = [ (inputs.den.namespace "eg" false) ];` and `imports = [ (inputs.den.namespace "eg" true) ];`? I believe the documentation uses `true`?

The boolean enables the flake output, use `false` for private namespaces. See [here](https://vic.github.io/den/namespaces.html)
 
>     * [ ]  is the full exposition exercised, i.e. the combination of `__findFile`, custom namespace and an `includes` with angle-bracket syntax?

I believe [it](https://github.com/vic/den/blob/main/templates/default/modules/namespace.nix) [is](https://github.com/vic/den/blob/main/templates/default/modules/aspects/defaults.nix).

 
>     * [ ]  if not, could [igloo.nix](https://github.com/vic/den/blob/f57da63114e2ccc2d465a6dc954192f0eb74d1b3/templates/default/modules/aspects/igloo.nix) potentially be a good place to rewrite into this direction to see if it really works in this kind of setup/situation?

Yep. See `defaults.nix` and `alice.nix` which both use __findFile. 
 
>     * [ ]  also, is there a (technical) possibility to have `__findFile` available automatically? however, this is obviously only a cosmetic improvement, but as far as I can see, apparently it is possible to have a custom namespace automatically in scope, so maybe yes?!

It is currently *experimental*, meaning I don't know how well it will play in the long run. That is why it is enabled explicitly by the user at the module level. I would even recommend not using it module-level (as a module arg) but instead a as a let-binding (see alice.nix)

vic:
I'm adding yet more checks to the default template regarding angle brackets.

Just wanted to say, thank you for using den <3, and discovering tests I'm missing, I'll add a couple just to make sure it must work for you.

xdbr:
I'm just genuinely happy to see such a great and well-thought out library/framework (i.e. also including flake-aspects, etc) in this problem space :) – in my opinion this is next-gen stuff and one of the best inventions since sliced bread (as we like to say around here). I'm really looking forward to checking out your latest developments first thing tomorrow morning. Keep cracking, I can't wait!

PR #80, opened 2025-11-12
default template includes routes.nix example and tests with namespaces and angle-brackets.

### LINKED ISSUE #78: outdated `_profile` examples.
they are using the old `userToHost`, etc. meaning that they are not being tested! test them or remove them.

vic:
This also removes outdated `_profile` directory. Closes #78. Instead our default template now includes an `eg/routes.nix` example router and exercises namespaces and angle-brackets, ensuring via tests that they work. 

cc @xdbr


xdbr:
@vic: To me, the new template looks just great! I'm currently putting it to work for my config and will keep you updated on how it goes. The only thing that I would like to add is that the creation of a VM and an xserver is a difficult assumption to make: in my case, for instance, I'm running the whole thing entirely on a headless server so I have to leave all of that out. Ideally, one would make that either optional or build a different template for that? What do you think?

BTW: This entire thing is still growing on me and it's such a huge pleasure! I keep notes on the side about different small things that I would like to share with you some time soon. So far I am the happiest person :)

vic:
Hey @xdbr, I'm so happy that you are happiest person!! haha. Sure, anything you want to share feel free, I'd like to know about your notes and what you have in mind. 

Yep, we could make the VM non-graphical easily, I'll keep that in mind. Please on an issue so I not forget, I believe it is something important. Also, currently I don't have access to a Mac (my old macbook died) so I don't know how VMs work in macs. Anyways, please open an issue for toggling or opt-out from graphical VM.

PR #117, opened 2025-12-11
feat: add inputs' and self' aspects

HeitorAugustoLN:
Adds `inputs'` and `self'` aspects

### DIFF
--- a/modules/aspects/provides/inputs.nix
+++ b/modules/aspects/provides/inputs.nix
@@ -1,0 +1,42 @@
+{ den, withSystem, ... }:
+{
+  den.provides.inputs' = den.lib.parametric {
+    description = ''
+      Provides the `flake-parts` `inputs'` (the flake's `inputs` with system pre-selected)
+      as a top-level module argument. This allows modules to access per-system
+      flake outputs without needing `pkgs.stdenv.hostPlatform.system`.
+
+      ## Usage
+
+          # makes inputs' available to modules in my-aspect
+          den.aspects.my-aspect.includes = [ den._.inputs' ];
+
+          # module implementation
+          { inputs', ... }: {
+            # use inputs' as needed
+          }
+    '';
+
+    includes = [
+      (
+        { host, user, ... }:
+        (withSystem host.system (
+          { inputs', ... }:
+          {
+            ${host.class}._module.args.inputs' = inputs';
+            ${user.class or null}._module.args.inputs' = inputs';
+          }
+        ))
+      )
+      (
+        { home, ... }:
+        (withSystem home.system (
+          { inputs', ... }:
+          {
+            ${home.class}._module.args.inputs' = inputs';
+          }
+        ))
+      )
+    ];
+  };
+}
--- a/modules/aspects/provides/self.nix
+++ b/modules/aspects/provides/self.nix
@@ -1,0 +1,42 @@
+{ den, withSystem, ... }:
+{
+  den.provides.self' = den.lib.parametric {
+    description = ''
+      Provides the `flake-parts` `self'` (the flake's `self` with system pre-selected)
+      as a top-level module argument. This allows modules to access per-system
+      flake outputs without needing `pkgs.stdenv.hostPlatform.system`.
+
+      ## Usage
+
+          # makes self' available to modules in my-aspect
+          den.aspects.my-aspect.includes = [ den._.self' ];
+
+          # module implementation
+          { self', ... }: { 
+            # use self' as needed
+          }
+    '';
+
+    includes = [
+      (
+        { host, user, ... }:
+        (withSystem host.system (
+          { self', ... }:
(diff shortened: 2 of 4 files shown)


vic:
Hey, thanks for the PR!

- I'm currently trying to decouple as much as possible from flake-parts and trying to reduce dependencies on it and its idioms. In the future I'd like den to be able to work without flakes and without flake-parts, so people using den will be able to choose if using flake-parts, or just evalModules like [dendritic-unflake](https://github.com/vic/dendritic-unflake) does or [falake](https://codeberg.org/FrdrCkII/falake). 

- What is the specific use for `self'` or `inputs'` in this PR ? I mean, if people can use the module-system to have access both of them why should Den include them as this PR does? 

I'm still working on decoupling flake-file from flakes (using [unflake](https://codeberg.org/goldstein/unflake)) but will keep this open and take a more deeper look later.

vic:
Den has a test suite now, and each feature is tested. Could you provide a test that shows it is NOT possible to use `inputs'`/`self'` unless we merge this PR ?

HeitorAugustoLN:
It is really not possible currently, but yeah, it won't be supported for configurations that aren't using flake-parts, so I will just use `inputs.input.packages.${host.system}.package`

vic:
Can you show me an example of where you are trying to use it and are unable to ? How about using `withSystem` instead ? 

We currently depend on `withSystem`, see https://github.com/vic/den/blob/3ec43279b5afea898737238a75178079bf91d606/modules/config.nix#L31

HeitorAugustoLN:
> Can you show me an example of where you are trying to use it and are unable to?

I can use it in perSystem or by using withSystem, so if I want to use it in an aspect I would have to use withSystem, but then it would be too verbose.

I currently have one use case that is basically importing my wrapped neovim from another flake with `home.packages = [ inputs'.neovim.packages.default ];`. since I haven't migrated yet to a dendritic approach, I use specialArgs in https://github.com/HeitorAugustoLN/nix-config/blob/1741a49777865cd1472ba3ad3d207b3dfe9c1abe/flake/configurations.nix#L139-L166

vic:
I haven't tried it but if `withSystem` gives you access to `inputs'` could we have a simple aspect that just exposes it ? 

I'd prefer if we made something like this possible:

```
{ withSystem, ... }: {
  den.aspects.provide-system-self = {
    _module.args.self' = (withSystem { self', ... }: self');
  };
}
```

But we extending a module that is included in all aspects, so all have access to this `self'` , what do you think ? 


In other words: being able to contribute to a module that is part of all aspects ? Is it `den.default` ?

modules/aspects/provides/inputs.nix
```diff
@@ -0,0 +1,41 @@
+{ den, withSystem, ... }:
+let
+  description = ''
+    Provides the `flake-parts` `inputs'` (the flake's inputs with system pre-selected)
```
vic: 👍 You are awesome, I like this a lot, `den.provides.inputs` is an opt-in feature for people who uses flake-parts.

vic: Not sure about the name, but I like it.

HeitorAugustoLN:
I am having some difficulty with the aspects lol, but I might have figured it out

modules/aspects/provides/self.nix
```diff
@@ -0,0 +1,47 @@
...
+          }
+    '';
+    __functor =
+      _:
+      let
+        inject =
+          context:
+          withSystem context.system (
+            { self', ... }:
+            {
```
vic: Try this: 

```
{ self', ... }:
{ class, ... }: {
  ${class}._module.args.self' = self';
}
```

~you wont need context anymore~.

HeitorAugustoLN:
Fixed the aspect issue, any thoughts on why it fails on darwin?

HeitorAugustoLN:
Ok, I understood what went wrong, the host and home we are checking is x86_64-linux only, when we run it through perSystem it uses its respective system such as aarch64-darwin, but the host and home still uses the same system as before

vic:
Awesome! I'm merging this!

Can I ask for another favor ? Please update the wiki documentation just so we have a way for people to discover `den._.inputs'` and `den._.self'`. 

Thanks!

HeitorAugustoLN:
Where is it?

vic:
Here: https://github.com/vic/den/wiki (it was private haha)

vic:
Get published on any file change at https://den.oeiuwq.com/

HeitorAugustoLN:
Ok!

HeitorAugustoLN:
I improved the descriptions of these aspects in #118, also I am not able to edit the wiki apparently

vic:
Oh, Ok. I'll update then wiki tomorrow then. Thanks so much I'll also merge #118

PR #119, opened 2025-12-13
add base module for user/host/home configs

vic:
Allows users to have a base module for home/host/user definitions:

```nix

den.base.host = {
  # applies to all den.hosts.<system>.<name>
  # use it to define options all hosts might have.
};

den.base.user = { ... } # included in all den.hosts.<system>.<name>.users.<user>

den.base.home = { ... } # included in all den.homes.<system>.<name>

den.base.conf = { ... } # included in ANY host / user / home config.
```

Checkout tests for a working example of this feature.

### DIFF
--- a/modules/_types.nix
+++ b/modules/_types.nix
@@ -26,6 +26,8 @@ let
       { name, config, ... }:
       {
         freeformType = lib.types.attrsOf lib.types.anything;
+        imports = [ den.base.host ];
+        config._module.args.host = config;
         options = {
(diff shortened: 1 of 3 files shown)


HeitorAugustoLN:
Really cool! I was a bit confused about this at first when reading the PR title, thinking it was basically `den.default`. But then I understood this is for the submodules for each definition of a host/user/home

vic:
Yep, this is cool. Feature was request by @CarlosZ at zulipchat.

PR #128, opened 2025-12-29
Refactor unfree aspect to use a module

Shawn8901:
This allows the aspect to be included in other aspects and compose all unfree packages that are defined.

:warning:  I just have tested the code for nixos and home-manager (as module) definitions.

### Context:
I want to use `<den/unfree>` in multiple aspects and have their defined packages to combined to one large `allowUnfreePredicate`.

That allows to define the unfree packages in the aspects near their usage and not on the respective class definition (e.G. the host).

Here an example on my tryout repo where i am currently exploring how to build my configs with den.

This defines an unfree predicate for vscode and the respective addons:

https://github.com/Shawn8901/dend_test/blob/main/modules/aspects/cfg/desktop/coding.nix

This defines my printer

https://github.com/Shawn8901/dend_test/blob/main/modules/aspects/cfg/printer.nix#L6

Both are later on then used in the host pointalpha together, whilst the printer aspect is not used on my zenbook.

### Possible workarounds

* Gather unfree packages by hand and define them once on a class
* use `nixpkgs.config.allowUnfree = true`

### Other resources

Discussion on zulip https://oeiuwq.zulipchat.com/#narrow/channel/548534-den/topic/feat-idea.20composable.20unfree.20batteries/with/564892231

Actual "root issue" why predicates are not combine/composable by default https://github.com/NixOS/nixpkgs/issues/197325

### Possible Todos

~One issue i found is, that it does not play well with setting~
```nix
 nixos.home-manager.useGlobalPkgs = true;
```
~as its setting a `nixpkgs.config.allowUnfreePredicate` even tho its not allowed when using `useGlobalPkgs` in homeManager class.~
~This should just be done if `useGlobalPkgs` is set to false, tho i am not yet fully confident how to set that condition correctly with den.~
Might have found a reasonable solution for that, see https://github.com/vic/den/pull/128#discussion_r2656546800

### DIFF
--- a/templates/examples/flake.lock
+++ b/templates/examples/flake.lock
@@ -22,11 +22,11 @@
     },
     "den": {
       "locked": {
-        "lastModified": 1763707606,
-        "narHash": "sha256-l9v3NNdKj3GJvV5LhzsWDs4Sl2bg0tuKNFFkMeFvUWo=",
+        "lastModified": 1767645236,
+        "narHash": "sha256-VMrWett3fWRb0hQh9K1JUC3zV2OlV4zOajFZHGr1GPw=",
         "owner": "vic",
         "repo": "den",
-        "rev": "8164e0d89c59839d67757bc9a1fb61770dc6e8b7",
+        "rev": "e7837daa0ea0f03fba62a2653ce16314a7d9d1c8",
         "type": "github"
       },
       "original": {
--- a/templates/examples/modules/_example/ci/unfree.nix
+++ b/templates/examples/modules/_example/ci/unfree.nix
@@ -1,6 +1,25 @@
 { den, ... }:
+
+let
+  codeAspect = {
+    includes = [ (den._.unfree [ "vscode" ]) ];
+    homeManager.programs.vscode.enable = true;
+  };
+  discordAspect = {
+    includes = [
+      (den._.unfree [ "discord" ])
+    ];
+    homeManager =
+      { pkgs, ... }:
+      {
+        home.packages = [ pkgs.discord ];
+      };
+  };
+in
 {
-  # cam uses unfree vscode.
-  den.aspects.cam.homeManager.programs.vscode.enable = true;
-  den.aspects.cam.includes = [ (den._.unfree [ "vscode" ]) ];
+  # cam uses unfree vscode and discord loaded from different aspects.
+  den.aspects.cam.includes = [
+    codeAspect
+    discordAspect
+  ];
 }


modules/aspects/provides/unfree/unfree-predicate-builder.nix
```diff
@@ -0,0 +1,89 @@
...
+          ${user.class}.imports = [
+            unfreeComposableModule
+            (
+              { config, osConfig, ... }:
+              {
+                nixpkgs.config.allowUnfreePredicate =
+                  if (!osConfig.home-manager.useGlobalPkgs) then
+                    (pkg: builtins.elem (lib.getName pkg) config.unfree.packages)
+                  else
+                    null;
```
Shawn8901: with that set, i try to avoid

```
evaluation warning: shawn profile: You have set either `nixpkgs.config` or `nixpkgs.overlays` while using `home-manager.useGlobalPkgs`.
                    This will soon not be possible. Please remove all `nixpkgs` options when using `home-manager.useGlobalPkgs`.
```

vic:
Hey, @Shawn8901 

Thanks a lot for this. All den features have tests in templates/examples/, see existing ones under ci/ directory. Could you please make sure we have at least a test that ensures the unfree aspect works as you intend ?

Shawn8901:
I have altered the unfree case to over the new changes. I hope that is okay, but as the whole logic now changes i thought that is a the best solution.

I had thought about adding extra checks to evaluate the unfreePredicate, tho if it would not compose the unfree package list it either should fail on building `vscode` or `discord` as then just one should be contained in the unfree package list.

Should i also add an extra case for having it as `nixosConfiguration`s?

vic:
your changes look good, updating the previous test is correct .

Shawn8901:
Thx for the feedback.
I'll check the ci failure later, today evening

Shawn8901:
> Thx for the feedback. I'll check the ci failure later, today evening

I hope the CI issue should be fixed now

modules/aspects/provides/unfree/unfree-predicate-builder.nix
```diff
@@ -0,0 +1,93 @@
...
+  unfreeComposableModule.options.unfree = {
+    packages = lib.mkOption {
+      type = lib.types.listOf lib.types.str;
+      default = [ ];
+    };
+  };
+
+  aspect = den.lib.parametric.exactly {
+    inherit description;
+    includes = [
```
vic, line 40: can we move these three included aspects to be let bindings so our includes array is short and references them ?

vic:
awesome work @Shawn8901!, looking good, just added a minor comment.

modules/aspects/provides/unfree/unfree-predicate-builder.nix
Shawn8901: Yeah, i hope its changed in the expected way :slightly_smiling_face:

Shawn8901:
let me inherit den.lib also in the prefix builder

edit: done

vic:
Thanks @Shawn8901

PR #132, opened 2026-01-28
Fix a single letter typo in vm.nix comment

nikolai-in:
Really important typo fix, probably the most important typo fix ever, all of my friends say so, so it must be true.


vic:
thank you!

PR #142, opened 2026-02-06
Simplify default template

vic:
This will help people getting started. 

Our old default template is now named example, as it provides some layout structure users can consider but it is not enforced at all.

Also renamed examples template to be CI now. since it is used for running all tests.


Buttars:
This example fails `nix flake check` with

```
den/templates/default on  main
❯ nix flake check
warning: unknown flake output 'modules'
evaluation warning: system.stateVersion is not set, defaulting to 26.05. Read why this matters on https://nixos.org/manual/nixos/stable/options.html#opt-system.stateVersion.
error:
       … while checking flake output 'nixosConfigurations'

       … while checking the NixOS configuration 'nixosConfigurations.igloo'

       … while calling the 'head' builtin
         at /nix/store/wzcf74b2g5c3qfg4rcswcz28ngiwzzp0-source/lib/attrsets.nix:1713:13:
         1712|           if length values == 1 || pred here (elemAt values 1) (head values) then
         1713|             head values
             |             ^
         1714|           else

       … while evaluating the option `system.build.toplevel':

       … while evaluating definitions from `/nix/store/wzcf74b2g5c3qfg4rcswcz28ngiwzzp0-source/nixos/modules/system/activation/top-level.nix':

       (stack trace truncated; use '--show-trace' to show the full, detailed trace)

       error:
       Failed assertions:
       - The ‘fileSystems’ option does not specify your root file system.
       - You must set the option ‘boot.loader.grub.devices’ or 'boot.loader.grub.mirroredBoots' to make the system bootable.

den/templates/default on  main
```

Buttars:
I added the following lines which seemed to fix the issue but I'm not sure what the implications are to the vm host. I'm not sure if this will affect iso builds.

```
    den.aspect.<vm host aspect>.nixos = {
      fileSystems."/".device = lib.mkDefault "/dev/noroot";
      boot.loader.grub.enable = lib.mkDefault false;
    };
```

vic:
Hey, @Buttars, thanks for reporting. 

The CI is green because test.yml creates a module having `_module.args.CI = {}` which enables the CI environment, and we have a `den._.ci-noboot` that when disables grub and root device when CI is present. 

We have to fix that, because as you mentioned, people are expecting to clone and have no problems. They don't have `_module.args.CI` and that's why there's no boot device. Feel free to submit a PR fixing this, I'm currently busy moving the context-propagation system outside of den.default.

vic:
grep the repo for `grub.enable`

PR #179, opened 2026-02-21
Remove OS from example defaults.nix

OscarMarshall:
Per the v0.9.0 upgrade instructions, taking `OS` as a parameter is no longer a thing. I noticed this instance when checking out the examples template to see how `den.ctx` affected it.


vic:
thanks I forgot that one when rewriting all tests.

Issue #195, opened 2026-02-24
FR: support unfree battery under `user` aspect


OscarMarshall:
Took some time to track this one down, but I figured it out! It's the new user class that's the issue here; not Home Manager this time.

Starting with 8d9cd1517cc0d414f8bfb938efd4eb587af4ed84 (user class, convenience for OS users.users.<userName> (https://github.com/vic/den/pull/170)), including `den._.unfree` under the user aspect results in an error:

```
error:
       … while evaluating the attribute 'value'
         at /nix/store/33dyb4wznryw1i0ypj5xw8bb70prirfa-source/lib/modules.nix:1160:7:
         1159|     // {
         1160|       value = addErrorContext "while evaluating the option `${showOption loc}':" value;
             |       ^
         1161|       inherit (res.defsFinal') highestPrio;

       … while evaluating the option `expr':

       … while evaluating the attribute 'mergedValue'
         at /nix/store/33dyb4wznryw1i0ypj5xw8bb70prirfa-source/lib/modules.nix:1234:5:
         1233|     # Type-check the remaining definitions, and merge them. Or throw if no definitions.
         1234|     mergedValue =
             |     ^
         1235|       if isDefined then

       … while evaluating definitions from `<unknown-file>':

       (stack trace truncated; use '--show-trace' to show the full, detailed trace)

       error: The option `users.users.tux.unfree' does not exist. Definition values:
       - In `/nix/store/vr2qkzcvhicjy60yvrzkqh9a7rr84997-source/modules/aspects/provides/unfree/unfree.nix, via option den.provides.unfree.<function body>.<function body>.user':
           {
             packages = [
               "vscode"
             ];
           }

       Did you mean `users.users.tux.enable', `users.users.tux.name' or `users.users.tux.expires'?
```

I've got a failing test which I'll open a PR for.

vic:
~This is a feature-request, to change the unfree battery to support aditional classes beyond nixos and homeManager.~

Indeed it was a bug :)

vic:
fixed by #196

PR #196, opened 2026-02-24
Only apply unfree to nixos/darwin/homeManager

OscarMarshall:
Here's my example failing test demonstrating #195.

### DIFF
--- a/modules/aspects/provides/unfree/unfree.nix
+++ b/modules/aspects/provides/unfree/unfree.nix
@@ -15,9 +15,20 @@ let
   __functor =
     _self: allowed-names:
     { class, aspect-chain }:
-    den.lib.take.unused aspect-chain {
-      ${class}.unfree.packages = allowed-names;
-    };
+    den.lib.take.unused aspect-chain (
+      if
+        (builtins.elem class [
+          "nixos"
+          "darwin"
+          "homeManager"
+        ])
+      then
+        {
+          ${class}.unfree.packages = allowed-names;
+        }
+      else
+        { }
+    );
 in
 {
   den.provides.unfree = {
--- a/modules/aspects/provides/unfree/unfree.nix
+++ b/modules/aspects/provides/unfree/unfree.nix
@@ -1,4 +1,3 @@
-{ den, ... }:
 let
   description = ''
     A class generic aspect that enables unfree packages by name.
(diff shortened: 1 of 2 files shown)


vic:
Oh, thanks for sending a proper test. :) 

This is what is happening:

When the `unfree` battery was made it had support only for nixos (os-level nixos.nixpks.config.unfreePredicate) and for homeManager given that the Home Manager project by itself also has an `nixpkgs.config`. 

The `user` forward class is trying to do set the `users.users.<userName>.unfree` option which does not exist.  Because `user` class is just a short-cut for nixos.users.users I believe the correct way would be to include your unfree definition at host level. 

It would be nice for `user`-class to also support this, right ? maybe sending a PR to the unfree battery ? or ask copilot to make a patch.

vic:
omg, test is passing now ?

OscarMarshall:
It only breaks when you try to realize `igloo.users.users.tux`; that's why the non-sensical `expr`. Otherwise, it's identical to `test-packages-set-on-home-manager` just above it.

vic:
oh my bad, I accepted copilot change. hha,

vic:
your code using user class was correctly showing the bug.

vic:
hm now I understand more (now I'm talking like an AI) the problem seems to be more about how forward battery works. because your original code was including the un free at user aspect level not at 'user' class, right.?

OscarMarshall:
Correct. I only mention the user _class_ because it introduced the bug.

modules/aspects/provides/unfree/unfree.nix
```diff
@@ -15,9 +15,20 @@ let
   __functor =
     _self: allowed-names:
     { class, aspect-chain }:
-    den.lib.take.unused aspect-chain {
-      ${class}.unfree.packages = allowed-names;
-    };
+    den.lib.take.unused aspect-chain (
```
vic: unused is not needed anymore, I disabled the deadnix checks for now. I did not like that we were poluting with these unused calls. please remove and simplify the code.

templates/ci/modules/features/batteries/unfree.nix
```diff
@@ -24,5 +24,17 @@
...
     );
 
+    test-user-class-works = denTest (
+      { den, igloo, ... }:
+      {
+        den.hosts.x86_64-linux.igloo.users.tux = { };
+        den.default.homeManager.home.stateVersion = "25.11";
+        den.aspects.tux.includes = [ (den._.unfree [ "vscode" ]) ];
+
+        expr = igloo.users.users.tux.name;
```
vic: Can you add a comment here? That we are forcing evaluation of users.users.tux.name just to check that the unfree battery does not adds anything inside users.users.tux.unfree ?

The comment is needed otherwise people seeing this in the future will think the code is not testing unfree properly. As happened to me.

vic: Perpaps even better, test that `!(igloo.users.users.tux ? unfree)`

OscarMarshall:
Do you squash PRs? Should I manually squash my commits? Any sort of commit style I should follow?

vic:
All merges are via squash dont worry. Thanks for all your work :)

PR #203, opened 2026-02-25
Change user shell assignment to `users.defaultUserShell`

drupol:
Let me know what you think. IMO, it looks simpler to use that.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -54,7 +54,7 @@ jobs:
     needs: [non-draft]
     name: approved
     runs-on: ubuntu-latest
-    if: ${{github.actor == github.repository_owner || github.event.review.state == 'approved'}}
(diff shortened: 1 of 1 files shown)


modules/aspects/provides/user-shell.nix
```diff
@@ -19,7 +19,7 @@ let
         { pkgs, ... }:
         {
           programs.${shell}.enable = true;
-          users.users.${user.userName}.shell = pkgs.${shell};
+          users.defaultUserShell = pkgs.${shell};
```
vic, line 22: But this will set for all users, instead of just for the particular user this battery might be enabled on.

vic:
perhaps another battery den._.default-shell might be better place for setting defaultUserShell.

modules/aspects/provides/user-shell.nix
drupol: Oooh true, sorry !

Issue #207, opened 2026-02-28
BUG: Nested provides must have `includes`


musjj:
### Description

Here's a reproduction repo: https://github.com/musjj/bogus-provides

Basically, an aspect within a `provides` can't be included if it doesn't have `includes`:

```nix
{
  inputs,
  den,
  foobar,
  ...
}:
{
  imports = [
    inputs.den.flakeModule
    (inputs.den.namespace "foobar" true)
  ];

  foobar.foo.provides.bar = den.lib.parametric { };

  den.hosts.x86_64-linux.igloo.users.tux = { };

  den.aspects.tux = {
    includes = [ foobar.foo._.bar ];
  };
}
```

The error:

```nix
error: attribute 'includes' missing
at /nix/store/za7gazmbfpn4zi3l1h5jdahy9yjlnspg-source/nix/lib.nix:38:73:
    37|         {
    38|           includes = map (applyStatics { inherit class aspect-chain; }) self.includes;
      |                                                                         ^
    39|         };
```

The error goes away once you've added an `includes` attribute to the aspect. The error doesn't happen with top-level aspects (e.g. `foobar.foo = den.lib.parametric { }`)

Is this the intended behavior?

vic:
not intented, maybe there are other places where we assume `self.includes` exist, we should at least fallback `self.includes or [ ]`, and need tests for these. please send PR fixing this.

musjj:
I'm unsure how to go about fixing this:

```nix
foobar.foo.provides.bar = den.lib.parametric { };

den.aspects.tux.includes = [
  (builtins.trace (builtins.attrNames foobar.foo) foobar.foo)
  (builtins.trace (builtins.attrNames foobar.foo._.bar) foobar.foo._.bar)
];
```

`foobar.foo` prints:

```nix
nix-unit-check> trace: [ "_" "__functor" "description" "includes" "modules" "name" "provides" "resolve" ]
```

`foobar.foo._.bar` prints:

```nix
nix-unit-check> trace: [ "__functionArgs" "__functor" ]
```

Which is supposed to be the correct behavior and what's influencing them?

This test is supposed to catch this issue:

https://github.com/vic/den/blob/fb189d3e83f76c56ee0d7b30752c9017fe4b5462/templates/ci/modules/features/deadbugs/_external-namespace-deep-aspect.nix#L14-L23

But they're excluded from the tests for some reason.

musjj:
Ok, it looks like that the issue only happens with `den.lib.parametric`:

[...]

Issue #113, opened 2025-11-22
BUG: unable to access any nested aspects from imported namespace


michaelBelsanti:
### Description
https://github.com/michaelBelsanti/bogus-den
```
       error: attribute '_' missing
       at /nix/store/1m2c8wls1ms8rzrfx37jdp9a3wvv066s-source/modules/bug.nix:7:32:
            6|   den.hosts.x86_64-linux.good.users.quasi = { };
            7|   den.aspects.bad.includes = [ styx.services._.ssh ];
             |                                ^
            8|   den.aspects.good.includes = [ styx.performance ];
```

These aspects work fine when being used within the original flake, but do not work when imported by another. Aspects that aren't nested work fine.

See: https://github.com/vic/den/discussions/112

michaelBelsanti:
This seems to still be an issue. Including a nested aspect is still throwing an error in the reproduction: https://github.com/michaelBelsanti/bogus-den/tree/namespace

```
nu ➜ nixos-rebuild-ng build --flake .#bad
building the system configuration...
error:
       … while calling the 'seq' builtin
         at /nix/store/ndig4f4dx8bmrmyr5vfm19g02r9l9ggm-source/lib/modules.nix:361:18:
          360|         options = checked options;
          361|         config = checked (removeAttrs config [ "_module" ]);
             |                  ^
          362|         _module = checked (config._module);

       … while evaluating a branch condition
         at /nix/store/ndig4f4dx8bmrmyr5vfm19g02r9l9ggm-source/lib/modules.nix:297:9:
          296|       checkUnmatched =
          297|         if config._module.check && config._module.freeformType == null && merged.unmatchedDefns != [ ] then
             |         ^
          298|           let

       … while evaluating the option `den.aspects.bad.__functor.<function body>.<function body>.__functor.<function body>.<function body>.includes':

       … while evaluating definitions from `<unknown-file>':

       … while evaluating the option `den.aspects.bad.__functor.<function body>.<function body>.includes':

       … while evaluating definitions from `/nix/store/qlg1nbcj59rsgxwf38m5lasxahshc6aj-source/modules/aspects/definition.nix':

       … while evaluating the option `den.aspects.bad.includes':

       … while evaluating definitions from `/nix/store/hsskrhb4f18rvf9wq7px1bf5wcsm4g7h-source/modules/bug.nix':

       (stack trace truncated; use '--show-trace' to show the full, detailed trace)

       error: attribute '_' missing
[...]

PR #208, opened 2026-03-01
fix include-less parametric in provides

musjj:
Fix: https://github.com/vic/den/issues/207

Also added a test to catch this.

I hope this isn't just peppering over a deeper underlying issue.

### DIFF
--- a/templates/ci/modules/features/provides-parametric.nix
+++ b/templates/ci/modules/features/provides-parametric.nix
@@ -19,7 +19,10 @@
         ns.bar.provides.baz = den.lib.parametric { };
         ns.a = den.lib.parametric {
           provides.b = den.lib.parametric { };
-          provides.c = den.lib.parametric { nixos.networking.hostName = "pinguino"; };
+          provides.c = den.lib.parametric { };
+          provides.d = den.lib.parametric {
+            provides.e = den.lib.parametric { nixos.networking.hostName = "pinguino"; };
+          };
         };
 
         den.aspects.igloo.includes = [
@@ -28,6 +31,8 @@
           ns.a
           ns.a._.b
           ns.a._.c
+          ns.a._.d
+          ns.a._.d._.e
         ];
 
         expr = igloo.networking.hostName;


musjj:
Never mind, this fails with:

```nix
ns.foo = den.lib.parametric { };
ns.bar.provides.baz = den.lib.parametric { };
ns.a = den.lib.parametric {
  provides.b = den.lib.parametric { };
  provides.c = den.lib.parametric { nixos.networking.hostName = "pinguino"; };
  provides.d = den.lib.parametric {
    provides.e = den.lib.parametric { };
  };
};

  ns.foo
  ns.bar._.baz
  ns.a
  ns.a._.b
  ns.a._.c
  ns.a._.d._.e
];
```

This errors with:

```nix
error: attribute '_' missing
at /nix/store/zs759xrvfvkk0aki4mpbdprshnjcmfc1-source/templates/ci/modules/features/provides-parametric.nix:34:11:
    33|           ns.a._.c
    34|           ns.a._.d._.e
      |           ^
    35|         ];
```

Looks like this is relevant: https://github.com/vic/den/issues/113

vic:
top level aspects always have an initial `includes = []` because they are coerced by the Nix module type defined at flake-aspect's types.nix. see https://github.com/vic/flake-aspects/blob/main/nix/types.nix#L81 , however doing `parametric { }` the argument to parametric here is a plain empty attr-set that was not coerced by the type-system.

musjj:
Hmm, interesting. Do you have any idea why the `provides` is yeeted out of the doubly-nested aspect?

```nix
ns.a = den.lib.parametric {
  provides.d = den.lib.parametric {
    provides.e = den.lib.parametric { };
  };
};
```

`ns.a._.d` works here, but `ns.a._.d._.e` fails.

Inspecting `ns.a._.d` reveals that it only has `__functionArgs` and `__functor`.

`provides`/`_` is entirely gone. Any idea why?

vic:
> Inspecting `ns.a._.d` reveals that it only has `__functionArgs` and `__functor`.

hm, the parametric.* function family should only change `__functor` and nothing more, maybe we need tests that ensure this, if you look at the implementation the reuse the aspect given to them and just change __functor. There's `lib.owned` that does remove __functor and includes. Maybe we need unit tests for them at ~/checkmate (as opposed to integration tests at templates/ci)

vic:
them having __functionArgs looks like already touched by Nix in some way. because we never add that.

musjj:
It's definitely not caused by the `parametric` functions. I tried debugging it like this:

```nix
parametric.__functor =
  _: aspect:
  let
    result = parametric.withOwn parametric.atLeast aspect;
  in
  (builtins.trace (builtins.attrNames result) result);
```

It prints:

```nix
trace: [ "__functor" "provides" ]
```

So the `provides` is still intact. 

I suspect that there was something wrong with the `provides` type: https://github.com/vic/flake-aspects/blob/88bac0fd89f3d9ce8f157debc57875ca470535e8/nix/types.nix#L84-L94, but it looks ok to me.

musjj:
Ok, I narrowed down the problem to the nested provide type being incorrectly coerced to `curriedProviderFn`. This is a very incomplete fix, but changing like this:

```diff
-curriedProviderFn = lib.types.functionTo providerType;
+curriedProviderFn = lib.types.addCheck (lib.types.functionTo providerType) (f: !(f ? provides));
```

Fixes the issue. Not sure what a more robust check should look like.

vic:
Right, looks like flake-aspects PR patch then, let's have tests there first.

musjj:
I tried to fix this here: https://github.com/vic/flake-aspects/pull/36

vic:
Merged your flake-aspects PR, do you think we need anything on Den itself ?

musjj:
Yeah the `self.includes or [ ]` is still needed. Also the parametric test is probably still useful to have here.

vic:
You are awesome, thanks for fixing all of this, and getting your feet wet on Den and flake-aspects internals :)

PR #221, opened 2026-03-04
fix wrong location issue of option definied in den.default

phhai-hcmut:
Set den.default as separate option with aspect type (similar to before context feature).


vic:
Thank you!

PR #236, opened 2026-03-06
feat(battery): New `den._.hostname` and .. 

musjj:
Convert two well-known patterns from the examples into built-in batteries:

- Bidirectional provider
- Automatically setting `networking.hostName`

I feel that these two use-cases are common enough that they are worth being lifted up into the built-in set of batteries.

What do you think?

Also, I would like to hear if you have any better ideas for the names:

- `den._.set-hostname`: Maybe too specific? In the future, will there be any other use-cases for auto-configuring hosts that could be included in this battery?
- `den._.bidirectional-provider`: Feels descriptive, but maybe too long?

### DIFF
--- a/docs/src/content/docs/guides/batteries.mdx
+++ b/docs/src/content/docs/guides/batteries.mdx
@@ -36,13 +36,13 @@ Sets the system hostname as defined in `den.hosts.<name>.hostName`:
 den.default.includes = [ den.provides.hostname ];
 ```
 
-### `den.provides.bidirectional-provider`
+### `den.provides.mutual-provider`
 
 Allows hosts and users to contribute configuration **to each other** through `provides`:
 
 ```nix
 den.hosts.x86_64-linux.igloo.users.tux = { };
-den.default.includes = [ den._.bidirectional-provider ];
+den.default.includes = [ den._.mutual-provider ];
 ```
 
 This is not the same as the built-in bidirectionality:
--- a/docs/src/content/docs/tutorials/example.md
+++ b/docs/src/content/docs/tutorials/example.md
@@ -3,7 +3,7 @@ title: "Template: Example"
 description: Feature showcase with namespaces, angle brackets, cross-platform aspects, and providers.
 ---
 
-The example template demonstrates Den's advanced features: namespaces, angle brackets, cross-platform hosts, bidirectional providers, and custom aspect libraries.
+The example template demonstrates Den's advanced features: namespaces, angle brackets, cross-platform hosts, mutual providers, and custom aspect libraries.
 
 ## Initialize
 
@@ -97,7 +97,7 @@ Aspects can have nested **provides** — `eg.vm._.gui` and `eg.vm._.tui` are sub
 
 The `<name>` syntax is shorthand for aspect lookup. `<den/define-user>` resolves to `den.provides.define-user`. See [Angle Brackets](/guides/angle-brackets/).
 
-### Bidirectional Providers
+### Mutual Providers
 
 ```nix
 # modules/aspects/alice.nix — user provides config TO the host
@@ -153,7 +153,7 @@ graph TD
 | Standalone HM config | ✓ |
 | Namespaces (`eg`) | ✓ |
 | Angle brackets | ✓ |
-| Bidirectional providers | ✓ |
+| Mutual providers | ✓ |
 | VM testing | ✓ |
 | CI checks | ✓ |
 
--- a/docs/src/content/docs/reference/batteries.mdx
+++ b/docs/src/content/docs/reference/batteries.mdx
@@ -38,7 +38,7 @@ Marks a user as the primary user (admin level) of a host. Works in NixOS/Darwin/
 Sets the user's login shell. 
 `users.users.<name>.shell` enabling the shell at host, and the home-manager `programs.<shell>.enable`.
 
-### `den._.bidirectional-provider`
+### `den._.mutual-provider`
 
 Allows the user and host to contribute configuration to each other via
 `provides`. This is not the same as the built-in bidirectionality. The
--- a/modules/aspects/provides/mutual-provider.nix
+++ b/modules/aspects/provides/mutual-provider.nix
@@ -33,7 +33,7 @@ let
     ## Usage
 
       den.hosts.x86_64-linux.igloo.users.tux = { };
-      den.default.includes = [ den._.bidirectional-provider ];
+      den.default.includes = [ den._.mutual-provider ];
 
     A user providing config TO the host:
 
@@ -55,7 +55,7 @@ let
   mutual = from: to: den.aspects.${from.aspect}._.${to.aspect} or { };
 in
 {
-  den.provides.bidirectional-provider =
+  den.provides.mutual-provider =
     { host, user }@ctx:
     den.lib.parametric.fixedTo ctx {
       inherit description;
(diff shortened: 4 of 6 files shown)


modules/aspects/provides/bidirectional-provider.nix
```diff
@@ -0,0 +1,55 @@
+# See usage at: defaults.nix, alice.nix, igloo.nix
+{ den, ... }:
+{
```
vic: Try using the same style as other batteires, use `let description = ...; in` and make the output attr short.

The style helps because you can define the function as a let binding and use take.exactly or parametric.exactly. 

```nix
let
  description = "Foo does bar";
  foo = { x }: { ... };
in 
den.provides.foo = take.exactly foo;
```

modules/aspects/provides/set-hostname.nix
```diff
@@ -0,0 +1,19 @@
+{ den, ... }:
+{
+  den.provides.set-hostname = den.lib.take.exactly (
+    { host }:
+    {
+      description = ''
+        Sets the system hostname as defined in `den.hosts`.
```
vic: as defined at `host.hostName`

in `den.hosts.x86_64-linux.igloo = {}`

`host.hostName` defaults to `"igloo"` 

You can explicitly change it using 

`den.hosts.x86_64-linux.igloo = { hostName = "iceberg"; }`

docs/src/content/docs/guides/batteries.mdx
```diff
@@ -28,6 +28,43 @@ den.default.includes = [ den.provides.define-user ];
 Sets `users.users.<name>` on NixOS/Darwin and `home.username`/`home.homeDirectory`
 for Home Manager. Works in both host-user and standalone home contexts.
 
+### `den.provides.set-hostname`
```
vic: maybe just `den.provides.hostname` is fine!

docs/src/content/docs/guides/batteries.mdx
```diff
@@ -28,6 +28,43 @@ den.default.includes = [ den.provides.define-user ];
...
+
+Automatically set the system host name to the one defined in `den.hosts`:
+
+```nix
+den.default.includes = [ den.provides.set-hostname ];
+```
+
+### `den.provides.bidirectional-provider`
```
vic: Perhaps making the name more explicit about host-user relationship, `den.provides.host-user-route` or something like that, open to suggestions.

docs/src/content/docs/tutorials/example.md
```diff
@@ -117,25 +117,6 @@ den.aspects.igloo = {
 
 Hosts and users can contribute configuration **to each other** through `provides`. Alice enables `nh` on igloo, and igloo enables `helix` for alice. These are cross-providers activated during context transformation.
 
-### Custom Routes
```
vic, line 120: The original intention of the routes example was to show people they can include aspects from other places.

I'd like to have some example for locating an aspect and including it just lie our `eg.routes` was doing.

Perhaps, something like, a host having 

```
den.x86_64-linux.hosts.igloo = {
  capabilities = {
    secure.allowedPorts = 22;
    ai.models = [ "opus" ];
  }
}
```

And then having an example that uses `host.capabilities.<cap-name>`

to locate `den.aspects.<cap-name>` and apply

```nix
includes = mapAttrsToList (
  name: args: 
   fixedTo (args // { host }) den.aspects.${name}
) host.capabilities
```

Something like that, showing people that you can include aspects from other namespace, here `den.aspects.<cap-name>`

vic:
Hey, @musjj. I like the idea of having them as batteries. I've added some comments for you.

docs/src/content/docs/tutorials/example.md
vic: Forget about this comment, these capabilities example is better suited for showing how to extend `den.ctx` (with those capabilities). 

Just remove the old code that showed our routes example.

vic:
Also, if we are having new batteries, be sure to also have some tests for them in templates/ci.

templates/ci/modules/features/deadbugs/cybolic-routes.nix
```diff
@@ -15,24 +15,7 @@
         den.default.homeManager.home.stateVersion = "25.11";
         den.hosts.x86_64-linux.igloo.users.tux = { };
 
-        den.aspects.routes =
```
vic, line 18: Keep this bug test as-is. Add another test for the battery.

docs/src/content/docs/guides/batteries.mdx
musjj: Not very explicit, but how about just `mutual-provider`? 

Tbh, I feel that the concept is difficult to express very succinctly.

`den` natively already has a concept of bidirectionality, so it's hard to differentiate the name while being concise.

modules/aspects/provides/bidirectional-provider.nix
```diff
@@ -0,0 +1,55 @@
+# See usage at: defaults.nix, alice.nix, igloo.nix
```
vic: Usage at templates/example/modules/{alice,igloo}.nix

modules/aspects/provides/bidirectional-provider.nix
```diff
@@ -0,0 +1,55 @@
...
+    Unlike `den.default` which is `parametric.atLeast` we use
+    `parametric.fixedTo` here, which help us propagate an already computed
+    context to all includes.
+
+    This battery, when installed in a `parametric.atLeast` will just forward
+    the same context.  The `mutual` helper returns an static configuration
+    which is ignored by parametric aspects, thus allowing non-existing
+    aspects to be just ignored.
+
+    Be sure to read: https://vic.github.io/den/dependencies.html
```
vic: That website does not exist anymore.

modules/aspects/provides/bidirectional-provider.nix
```diff
@@ -0,0 +1,55 @@
+# See usage at: defaults.nix, alice.nix, igloo.nix
+{ den, ... }:
+let
+  description = ''
+    Allows hosts and users to contribute configuration **to each other**
+    through `provides`.
```
vic: Mention that Den is already bidirectional.

```
# contributes to ALL users of this host
den.aspects.my-host.homeManager = { ... }

# contributes to ALL hosts of where my-user exist
den.aspects.my-user.nixos = { ... }
```

This battery is intended for per-user/per-host specific settings. via

```
den.aspects.my-user.provides.my-host.nixos = { ... }
```

be sure to also mention this difference in the docs.

modules/aspects/provides/bidirectional-provider.nix
musjj: Is there an equivalent page for the current website?

vic: https://den.oeiuwq.com/explanation/context-pipeline/ the section about Host Context.

docs/src/content/docs/guides/batteries.mdx
vic: Yeah, lets name it `mutual-provider` and leave the bi-directionality term for talking about Den core.

vic:
Looks good, just rename to `mutual-provider`, update the link and we'll merge.

musjj:
Done!

vic:
thanks!

PR #244, opened 2026-03-09
fix(battery): den._.hostname strictness

vic:
This changes `den._.hostname` from `take.exactly {host}` to `take.atLeast {host,...}`, to make it also work on user aspects because they have no `{host}` context, only a `{host, user}` one.

Also added tests that including `den._.hostname` in a second order include works. NOTE: the aspect using `includes = [ den._.hostname ]` needs to be `parametric {}` to be able to forward context.

Might be of interest @drupol

### DIFF
--- a/modules/aspects/provides/hostname.nix
+++ b/modules/aspects/provides/hostname.nix
@@ -11,13 +11,13 @@ let
   '';
 
   setHostname =
-    { host }:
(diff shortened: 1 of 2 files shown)


drupol:
Testing now.

drupol:
Replied on Zulip!

Issue #263, opened 2026-03-11
Fix npins instructions to use main branches like flakes do.


ulfbayte:
Hello,

First, thanks for all of this and also making a tutorial :)

I'm a beginner and I'm doing the tutorial https://den.oeiuwq.com/guides/from-zero-to-den/#step-2-pinning-den-and-dependencies so I can't use the standard process for reporting bugs.

I hope I don't waste anyone time with this...

I'm just doing the step 2 

<img width="764" height="336" alt="Image" src="https://github.com/user-attachments/assets/ac237a53-73d3-4ef1-aca2-26a9542c16a9" />

The npins + home manager part isn't happy.
Error message seems legit but I think I had to report it here.

```console
[ulfbayte@cyrix:~/nix-config]$ l
total 12K
drwxr-xr-x  3 ulfbayte users 4,0K 11 mars  21:40 .
drwx------ 20 ulfbayte users 4,0K 11 mars  21:25 ..
drwxr-xr-x  2 ulfbayte users 4,0K 11 mars  21:17 modules

[ulfbayte@cyrix:~/nix-config]$ cp -r /etc/nixos/ modules/_nixos

[ulfbayte@cyrix:~/nix-config]$ npins init
[INFO ] Welcome to npins!
[INFO ] Creating `npins` directory
[INFO ] Writing default.nix
[INFO ] Writing initial lock file with nixpkgs entry (need to fetch latest commit first)
[INFO ] Successfully written initial files to 'npins/sources.json'.

[ulfbayte@cyrix:~/nix-config]$ npins add github vic import-tree
[INFO ] Adding 'import-tree' …
    repository: https://github.com/vic/import-tree.git
    pre_releases: false
    submodules: false
    version: v0.1.0
    revision: a037ed2a58fc0ebed9e93b9ef79b0646e648f719
    hash: sha256-TEbh9zISiQcU82VwVoEbmXHnSGlUxTwvjJA9g9ErSDA=
    frozen: false


[ulfbayte@cyrix:~/nix-config]$ npins add github vic flake-aspects
[INFO ] Adding 'flake-aspects' …
    repository: https://github.com/vic/flake-aspects.git
    pre_releases: false
    submodules: false
    version: v0.6.0
    revision: 81a51a8997abe392b9d0794424a4823adc9bd3af
    hash: sha256-e3YqMwdQ2EpVTuWRkK2rkrHywp8QCpkj7x2U53SNoIA=
    frozen: false


[...]

PR #264, opened 2026-03-11
Update home-manager command to specify master branch

ulfbayte:
FIX : https://github.com/vic/den/issues/263
home-manager does not have any release tags


vic:
Thank you so much! <3

PR #291, opened 2026-03-17
docs: clarify how to enable angle-brackets in a file

crasm:


vic:
much better, thank you :)

PR #293, opened 2026-03-17
fix(ctx): should not use hm from host when no-bidir enabled.

### LINKED ISSUE #292: homeManager values from host used even when bidirectionality is disabled.
[#den > duplication issue](https://oeiuwq.zulipchat.com/#narrow/channel/548534-den/topic/duplication.20issue/with/579774482)

vic:
cc @Sharparam, offending line was: https://github.com/vic/den/pull/293/changes#diff-a008f96c95a197db435b512a966b1b8a2038e6da23230fddd6184e120aa22b15L75

Fixes #292.

### DIFF
--- a/nix/lib/home-env.nix
+++ b/nix/lib/home-env.nix
@@ -72,7 +72,6 @@ let
       includes = [
         (den.ctx."${ctxName}-user" { inherit host user; })
         (den.ctx.user { inherit host user; })
-        (den.lib.parametric.fixedTo { inherit host user; } den.aspects.${host.aspect})
       ];
     };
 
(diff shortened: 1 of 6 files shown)


nix/lib/home-env.nix
```diff
@@ -72,7 +72,6 @@ let
       includes = [
         (den.ctx."${ctxName}-user" { inherit host user; })
         (den.ctx.user { inherit host user; })
-        (den.lib.parametric.fixedTo { inherit host user; } den.aspects.${host.aspect})
```
vic, line 75: This was unconditionally reading from the host aspect

PR #295, opened 2026-03-17
doc: expand on docs contribution instructions

crasm:


crasm:
Hmm, I suggest changing the semantic PR check from `doc` to `docs` as allowed prefix

PR #303, opened 2026-03-18
feat(lib): add den.lib.perHome and add documentation for context shortcuts

HeitorAugustoLN:
Noticed these missing from #288, so decided to add it


vic:
good catch! thanks

PR #323, opened 2026-03-20
docs: host-specific standalone homes

musjj:
Add documentation for the host-specific standalone home feature.


vic:
Awesome, thanks!

PR #340, opened 2026-03-27
feat(core): add fixedTo.{atLeast,exactly,upTo}

DylanRJohnston:
## Overview
Adds new features, early draft, wanted to get your opinion on the design before committing to cleaning it up and potentially breaking it into smaller focused PRs. Needs documentation update too.

#### `den.lib.take.upTo`
Uses the same predicate as `atLeast` to decide whether to invoke a function, but calls the function with only the arguments it supports unlike `atLeast` which will call the function with extra arguments causing an error unless the function signature uses `...`. In my opinion this is how `atLeast` should function, but I understand that this is a breaking change if anyone is using `args@{ ctxKey, ... }` to capture the full context. So I've introduced this as a seperate parametric type.

#### `den.lib.canTake.upTo`
An alias of `den.lib.canTake.atLeast` provided for API symmetry between `den.lib.canTake` and `den.lib.take`

#### `den.lib.recursiveFunctor`
Much like `den.lib.functor` this takes an `apply` function and an aspect and applies it to each include in the aspects `includes` list. However, this also recurses into the include list and applies the function down the include tree. Much like `fixedTo` does but without owned or static aspects.

*Caveat:* This currently doesn't resolve static or parametric includes and then recurse into their includes if they exist, should it? I'm not sure why you would write aspects with this configuration, but it's theoretically possible. I originally didn't implement this because I thought it would be too complicated, but I think you can just call the aspects with ctx and static ctx and then recurse into them to get at them?

#### `den.lib.parametric.fixedTo.{exactly,atLeast,upTo}`
This family of `fixedTo` variants, follow `fixedTo`'s ability to recurse into the import tree, but do not evaluate owned or static aspects. This reflects the relationship between `parametric.__functor` and `parametric.{exactly,atLeast,upTo}`.

### DIFF
--- a/nix/lib/parametric.nix
+++ b/nix/lib/parametric.nix
@@ -52,10 +52,10 @@ let
 
   includeNothing = (_: _: { });
 
-  parametric.deepOwned = functor: deepRecurse includeOwnedAndStatics functor lib.id;
+  parametric.deep = functor: deepRecurse includeOwnedAndStatics functor lib.id;
   parametric.deepParametrics = functor: deepRecurse includeNothing lib.id functor;
 
-  parametric.fixedTo.__functor = _: attrs: parametric.deepOwned (lib.flip parametric.atLeast attrs);
+  parametric.fixedTo.__functor = _: attrs: parametric.deep (lib.flip parametric.atLeast attrs);
   parametric.fixedTo.exactly = attrs: parametric.deepParametrics (lib.flip take.exactly attrs);
   parametric.fixedTo.atLeast = attrs: parametric.deepParametrics (lib.flip take.atLeast attrs);
   parametric.fixedTo.upTo = attrs: parametric.deepParametrics (lib.flip take.upTo attrs);


templates/ci/modules/features/parametric.nix
```diff
@@ -113,5 +136,145 @@
       }
     );
 
+    test-parametric-fixedTo-atLeast = denTest (
```
DylanRJohnston: This is a giant test of all the new functionality, I would need to split this apart.

templates/ci/modules/features/parametric.nix
```diff
@@ -113,5 +136,145 @@
...
+          }
+          |> lib.mapAttrs (
+            _: test:
+            den.aspects.outer
+            |> testOptionProvider (builtins.attrNames test.ctx)
+            |> (lib.flip test.functor) test.ctx
+            |> den.lib.aspects.resolve "nixos" [ ]
+            |> (nixos: inputs.nixpkgs.lib.evalModules { modules = [ nixos ]; })
+            |> (x: x.config.test)
+          );
```
DylanRJohnston: When I was working on this, I wanted to stay away from `ctx.host` and `den.host` because they seemed the most magical to me at the time and I wanted to ensure if I hit any problems it was with this specifically, so there's some machinery here that would normally be handled by `den.host`. Should be fairly simple to switch it over to be consistent with the rest of the tests.

vic:
Hey @DylanRJohnston you are awesome, I like this very much!

Indeed, adding all these parametrics makes sense, thanks for taking the time to make this PR. 

Some notes:

- Today I merged #335 and it was big, so this branch has conflicts. I recommend splitting this PR. You will notice I tried to start some `parametric.deep <functor>` combinator but actually never finished it, feel free to refactor/remove it because your `recursiveFunctor` is better. As long as our current tests are green you are free to move our impl. 

- We don't use `|>` pipe operator, since that needs people to enable an experimental feature and we can't rely on non-standard features. use `lib.pipe`. 

- Yes, please split tests, use sub-directories if you like, naming is important `flake.tests.<suite>` so we can run `just ci <suite>`, other than that, feel free to split large tests into their own modules.


- It is ok if tests use custom ctx, `denTest` are isolated and it is actually good we can use them for testing how parametric families behave.

I won't have internet today and maybe also not tomorrow. (My dad is having a surgery today and I'll be at the hospital, and tomorrow I get baptized at church -- so maybe I can get back to you on this until sunday).

Thanks for all your effort on this :)

Btw, if you can experiment and switch our default functor to be `upTo` and tests pass feel free. :)

templates/ci/modules/features/parametric.nix
```diff
@@ -113,5 +136,145 @@
...
+            };
+
+          _.static =
+            { class, ... }:
+            {
+              ${class}.test = [ "excluded-static-${name}" ];
+            };
+
+          includes = include ++ [
+            den.aspects.${name}._.host
```
vic: I belive there's a fixed point for aspects: { aspect,...} see lib/aspects/types.nix so you don't have to access by name, aspect can reference itself, just an option.

templates/ci/modules/features/parametric.nix
```diff
@@ -113,5 +136,145 @@
...
+            #   };
+            #   parametricFunctor = atLeast.fixed;
+            # };
+          }
+          |> lib.mapAttrs (
+            _: test:
+            den.aspects.outer
+            |> testOptionProvider (builtins.attrNames test.ctx)
+            |> (lib.flip test.functor) test.ctx
+            |> den.lib.aspects.resolve "nixos" [ ]
```
vic: this api changed when we bundled flake aspects, no need for the list at the end

templates/ci/modules/features/parametric.nix
```diff
@@ -113,5 +136,145 @@
...
+            |> (nixos: inputs.nixpkgs.lib.evalModules { modules = [ nixos ]; })
```
vic: how about defining options.test here, since it is used for checking values. instead of defining the option at an aspect.

templates/ci/modules/features/parametric.nix
```diff
@@ -113,5 +136,145 @@
...
+            |> (x: x.config.test)
+          );
+
+        expected = {
+          atLeastHost = [
+            "igloo-inner"
```
vic: I'd like to have these assertions separated. you can still use a factory function that wraps denTest, the reason is we better have independent tests that can give us clues of what to fix in the future.

vic (changes requested):

vic:
Hey @DylanRJohnston, I'm trying to make some progress on this, starting with merging `upTo` at #384

nix/lib/can-take.nix
```diff
@@ -8,10 +8,10 @@ let
       required = builtins.filter (n: !args.${n}) (builtins.attrNames args);
       intersect = builtins.intersectAttrs args params;
     in
-    {
+    rec {
       satisfied = valid && builtins.all (n: params ? ${n}) required;
       exactly = valid && required == builtins.attrNames params;
-      upTo = valid && intersect != { };
+      upTo = satisfied && intersect != { };
```
DylanRJohnston, line 15: `upTo` shouldn't return true for functions that require more than you're able to provide, the previous implementation returned true for `canTake { host = true } ({ host, user }: {})`

nix/lib/parametric.nix
```diff
@@ -12,17 +12,13 @@ let
...
       includes = map (
-        include:
-        if include ? includes then
-          if include ? __functor then include else parametric.deep functor ctx include
-        else
-          include
-      ) (provided.includes or [ ]);
+        include: if include ? includes && !include ? __functor then branch include else leaf include
+      ) (aspect.includes or [ ]);
     };
```
DylanRJohnston, line 22: The name change comes from the fact that this function isn't recursive anymore, it just applies two different functions to one level of the include tree, one for branches, and one for leaves.

nix/lib/parametric.nix
```diff
@@ -32,23 +28,37 @@ let
...
+
+  includeNothing = (_: _: { });
+
+  parametric.deepOwned = functor: deepRecurse includeOwnedAndStatics functor lib.id;
+  parametric.deepParametrics = functor: deepRecurse includeNothing lib.id functor;
+
+  parametric.fixedTo.__functor = _: attrs: parametric.deepOwned (lib.flip parametric.atLeast attrs);
+  parametric.fixedTo.exactly = attrs: parametric.deepParametrics (lib.flip take.exactly attrs);
+  parametric.fixedTo.atLeast = attrs: parametric.deepParametrics (lib.flip take.atLeast attrs);
+  parametric.fixedTo.upTo = attrs: parametric.deepParametrics (lib.flip take.upTo attrs);
```
DylanRJohnston: When implementations unify it's usually a good sign you're on the right track. I found I was able to unify the recursion in my implementation with the recursion in `parametric.deep`. Interestingly one applies the functor on the node in the include tree before it gets traversed, the other applies the functor at the leaf nodes. Not 100% sure why the difference is required or if one could be massaged into the other.

I found it particularly interesting that `attrs` / `ctx` could be factored out right to the top level and is no longer interleaved through the include tree traversal but instead is only partially applied to the functor.

nix/lib/can-take.nix
vic: Good catch, I have some usage of upTo in den core, lets see if tests pass.

nix/lib/parametric.nix
vic: +1

nix/lib/can-take.nix
```diff
@@ -8,10 +8,10 @@ let
       required = builtins.filter (n: !args.${n}) (builtins.attrNames args);
       intersect = builtins.intersectAttrs args params;
     in
-    {
+    rec {
```
vic: This is the first `rec` in codebase, can we avoid it?

templates/ci/modules/features/parametric-fixedTo.nix
```diff
@@ -0,0 +1,459 @@
...
+
+            den.ctx.host.includes = [
+              test-option
+            ];
+
+            den.ctx.user.includes = [
+              ({ host, user }: den.lib.parametric.fixedTo.upTo { inherit host user; } (test-aspect aspects))
+            ];
+
+            expr = lib.sort (a: b: a < b) igloo.test;
```
vic: we have been using `lib.sort lib.lessThan` in some places.

templates/ci/modules/features/parametric.nix
vic: haha yes please.

templates/ci/modules/features/parametric-fixedTo.nix
DylanRJohnston: Yeah when I was writing the tests I wanted to avoid depending on the internal ordering of the module resolution

vic:
Awesome @DylanRJohnston everything is green, shall we merge ?

DylanRJohnston:
> Awesome @DylanRJohnston everything is green, shall we merge ?
Yeah so long as you're happy with the names of everything

vic:
I like much more the names you use than mine :D. I will merge, infinite thanks! 

Hopefully later we can make your wish come true, make upTo the default parametric.

PR #342, opened 2026-03-28
feat(core): support functionArgs in IntoPath

Damnjelly:
Adds support for functionArgs inside the IntoPath arg of den.provides.forward. 
Includes tests for both home manager and nixos (might be a bit overkill?).
Also fixes a typo in the documentation `permanance -> permanence`

**usage example:**
```nix
  sysPersist =
    { host }:
    {
      class,
      aspect-chain,
    }:
    den._.forward {
      each = lib.singleton class;
      fromClass = _: "sysPersist";
      intoClass = _: host.class;
      intoPath =
        _:
        { config, ... }:
        [
          "environment"
          "persistence"
          config.impermanence.persistence-dir
        ];
      fromAspect = _: lib.head aspect-chain;
      guard = { options, ... }: options ? environment.persistence;
    };
```

### DIFF
--- a/modules/aspects/provides/forward.nix
+++ b/modules/aspects/provides/forward.nix
@@ -49,15 +49,11 @@ let
     let
       fromClass = fwd.fromClass item;
       intoClass = fwd.intoClass item;
-      intoPath =
-        let
-          raw = fwd.intoPath;
-        in
-        if lib.functionArgs raw != { } then raw else raw item;
+      intoPath = fwd.intoPath item;
 
-      intoPathArgs = if builtins.isFunction intoPath then lib.functionArgs intoPath else { };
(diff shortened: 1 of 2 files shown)


modules/aspects/provides/forward.nix
```diff
@@ -49,7 +49,15 @@ let
     let
       fromClass = fwd.fromClass item;
       intoClass = fwd.intoClass item;
-      intoPath = fwd.intoPath item;
+      intoPath =
```
vic: what if people do need to access item and the module args. `item: args:` can you change that? if (intoPath item) returns a function apply module args to it

vic:
Thanks @Damnjelly.

PR #347, opened 2026-03-30
fix: `nh` invocation for standalone homes

theutz:
Checks if the command is `home`, and if so, properly formats the arguments so that `nh` will be able to switch to the specified home.

Fixes #253


vic:
Thank you very much, @theutz !

PR #353, opened 2026-04-01
fix(namespaces): fix missing reflection information on imported providers from a namespace

### LINKED ISSUE #352: BUG: Exported providers lose function argument reflection
### Description
Providers exported through a namespace lose their function argument reflection upon import breaking ctx invocation (they get called with `{class, aspect-chain}` because aspect resolution thinks they're static aspects). You can see the divergence in behaviour of namespace exported providers between `test-internal-exported-providers-function-arg-reflection` and `test-external-exported-providers-function-arg-reflection` where the later fails with `{ }`. I wasn't sure how best to simulate a namespace import in a test so I went the perhaps overkill method of using `inputs.nixpkgs.lib.evalModules`.

```nix
{ denTest, ... }:
{
  flake.tests.deadbugs-issue-xxx =
    let
      module =
        { inputs, ... }:
        {
          imports = [
            inputs.den.flakeModule
            (inputs.den.namespace "test" true)
          ];

          test.aspect._.host =
            { host, ... }:
            {
              nixos.environment.sessionVariables.TEST_HOST = host.name;
            };
        };
      internal = module;
      external =
        inputs:
        (inputs.nixpkgs.lib.evalModules {
          specialArgs = { inherit inputs; };
          modules = [ module ];
        }).config.flake;
    in
    {
      test-internal-exported-providers-function-arg-reflection = denTest (
        {
          lib,
          test,
          ...
        }:
        {
          imports = [ internal ];

          expr = lib.functionArgs test.aspect._.host;
          expected = {
            host = false;
          };
        }
      );

      test-raw-exported-providers-function-arg-reflection = denTest (
        {
          inputs,
          lib,
          test,
          ...
        }:
        {
          expr = lib.functionArgs (external inputs).denful.test.aspect._.host;
          expected = {
            host = false;
          };
        }
      );

      test-external-exported-providers-function-arg-reflection = denTest (
        {
          inputs,
          lib,
          test,
          ...
        }:

        {
          imports = [ (inputs.den.namespace "test" [ (external inputs) ]) ];
          expr = lib.functionArgs test.aspect._.host;
          expected = {
            host = false;
          };
        }
      );
    };
}
```

DylanRJohnston:
Fixes #352

## Summary
Providers when imported through a namespace get pass through a functor transformation in `nix/lib/namespace/nix:functorModules` which fails to propagate its `__functionArgs` for the reflection needing in the context propagation system. Adding this alone is not enough because then this triggers the freeformType part of the aspect type definition resulting in a `__functionArgs` like this. The fix as far as I can tell also requires registering the `__functionArgs` type with the aspect type.

```nix
 __functionArgs = {
        imports = [
          {
            _file = "<unknown-file>, via option den.ful.test.aspect.provides.host.__functionArgs";
            imports = [
              { host = false; }
            ];
          }
        ];
      };
 ```

### DIFF
--- a/nix/lib/aspects/types.nix
+++ b/nix/lib/aspects/types.nix
@@ -26,8 +26,7 @@ let
     in
     names != [ ] && builtins.all (n: builtins.elem n providerArgNames) names;
 
-  directProviderFn =
-    cnf: lib.types.addCheck (lastFunctionTo (aspectSubmodule cnf)) isProviderFn;
+  directProviderFn = cnf: lib.types.addCheck (lastFunctionTo (aspectSubmodule cnf)) isProviderFn;
 
   curriedProviderFn =
     cnf:
--- a/README.md
+++ b/README.md
@@ -16,8 +16,6 @@
 These configurations become specific when applied to your particular infra entities (hosts/users),
 while allowing re-usable aspects to be shared between hosts, users, or across other flakes and non-flake projects.
 
-
-
 <table>
 <tr>
 <td>
--- a/nix/lib/parametric.nix
+++ b/nix/lib/parametric.nix
@@ -19,10 +19,7 @@ let
       includes = map (
         include:
         if include ? includes then
-          if include ? __functor then
-            include
-          else
-            parametric.deep functor ctx include
+          if include ? __functor then include else parametric.deep functor ctx include
         else
           include
       ) (provided.includes or [ ]);

### CODE REVIEW

File: nix/lib/namespace.nix [hunk 1]
```diff
@@ -36,7 +36,10 @@ let
     aspectPath: v:
     lib.optionals (builtins.isAttrs v) (
       lib.optional (v ? __functor) {
-        config = lib.setAttrByPath aspectPath { __functor = v.__functor; };
+        config = lib.setAttrByPath aspectPath {
+          __functor = v.__functor;
+          __functionArgs = v.__functionArgs or { };
```
DylanRJohnston: Is it better to not set `__functionArgs` at all if the incoming functor lacks that information? Or is setting it to `{ }` the same?

nix/lib/aspects/types.nix
```diff
@@ -112,6 +112,14 @@ let
...
             default = cnf.defaultFunctor or lib.const;
           };
+
+          __functionArgs = lib.mkOption {
+            internal = true;
+            visible = false;
+            description = "Argument reflection information for the functor";
+            type = lib.types.lazyAttrsOf lib.types.bool;
+            default = { };
+          };
```
DylanRJohnston: This is required to avoid triggering the freeformType (I think?) and making it end up like

```nix
 __functionArgs = {
        imports = [
          {
            _file = "<unknown-file>, via option den.ful.test.aspect.provides.host.__functionArgs";
            imports = [
              { host = false; }
            ];
          }
        ];
      };
```

nix/lib/namespace.nix
vic: I don't think setting to `{ }` is the same as not having `__functionArgs`, having the empty attrset is like saying the function takes no args at all.

DylanRJohnston: Or it takes ` { ... }` I think? I'll change it, is there an easy way to conditionally add an attribute to an attribute set?

vic: Not pretty but I'd go for `{ ... } // (lib.optionalAttrs predicate { ... })`

nix/lib/aspects/types.nix
```diff
@@ -112,6 +112,14 @@ let
             type = functorType;
             default = cnf.defaultFunctor or lib.const;
           };
+
+          __functionArgs = lib.mkOption {
+            internal = true;
+            visible = false;
+            description = "Argument reflection information for the functor";
+            type = lib.types.lazyAttrsOf lib.types.bool;
+            default = { };
```
vic: This default value worries me a bit, try this in a repl: 

```nix
lib.functionArgs { __functor = _: { x }: x; __functionArgs = { y = false; }; }
```

Maybe the default can be `default = lib.functionArgs (config.__functor config)` ?

DylanRJohnston:
Can I ask what's the correct way to run the tests in `templates/ci` against local changes to the lib? I ended up modifying `flake.nix` to point `inputs.den` at `path:../..`, but it's a bit annoying juggling that edit around and not accidentally committing it.

vic:
> what's the correct way to run the tests in `templates/ci`

Use `just ci`

nix/lib/aspects/types.nix
DylanRJohnston: That causes the following test failure

```
☢️ standalone-homes.test-home-standalone-without-existing-host
error:
       … while evaluating the attribute 'optionalValue.value'
         at /nix/store/660ljssgj5ky0ph0z8pwc1jw8q39xb6j-source/lib/modules.nix:1298:41:
         1297|
         1298|     optionalValue = if isDefined then { value = mergedValue; } else { };
             |                                         ^
         1299|   };

       … while evaluating the attribute 'value'
         at /nix/store/660ljssgj5ky0ph0z8pwc1jw8q39xb6j-source/lib/types.nix:879:13:
          878|             headError = checkDefsForError check loc defs;
          879|             value = mapAttrs (
             |             ^
          880|               n: v:

       … while evaluating definitions from `/nix/store/pyrj41bjak67a01dfkd1f8vy2crlfvrr-source/templates/ci/modules/features/homes.nix':

       (stack trace truncated; use '--show-trace' to show the full, detailed trace)

       error: function '__functor' called without required argument 'class'
       at /nix/store/pyrj41bjak67a01dfkd1f8vy2crlfvrr-source/nix/lib/parametric.nix:37:9:
           36|         self:
           37|         { class, aspect-chain }:
             |         ^
           38|         {
```

DylanRJohnston: Actually that fails for me with `default = { };` as well 🤔

vic: `default = lib.functionArgs config` should just work, tried at repl: `lib.functionArgs { __functor = x: { y }: y; }`

DylanRJohnston: It causes an infinite recursion unfortunately

vic: I'm just worried that having that `__functionArgs` by default will mask the real arguments of the functor.

Can we solve this without having the `__functionArgs` option ?

DylanRJohnston: It gets mangled into this by the freeformType deferred modules type

```
 __functionArgs = {
        imports = [
          {
            _file = "<unknown-file>, via option den.ful.test.aspect.provides.host.__functionArgs";
            imports = [
              { host = false; }
            ];
          }
        ];
      };
```

DylanRJohnston: Alright what about this approach of special casing the freeformType

```nix
freeformType = lib.types.mkOptionType {
  name = "Aspect Freeform Type";
  merge =
    loc: defs:
    let
      functionArgs = lib.pipe defs [
        (lib.filter (def: def.value ? __functionArgs))
        (lib.map (def: def.value.__functionArgs))
        (x: if x != [ ] then lib.last x else null)
      ];
      everythingElse = lib.filter (def: !def.value ? __functionArgs) defs;
    in
    (lib.types.lazyAttrsOf lib.types.deferredModule).merge loc everythingElse
    // lib.optionalAttrs (functionArgs != null) {
      __functionArgs = functionArgs;
    };
};
```

vic: how about specifying functionArgs in a nested functor?

```
__functor = self: {
  __functor = realFunctor;
  __functionArgs = { x = false; };
}
```

from repl:

```console
nix-repl> lib.functionArgs { __functor = x: { __functor = y: { z }: z; __functionArgs = { w = false; }; }; }
{ w = false; }
```

Oh but the problem is we are loosing the `aspect.__functor` args, right? And using `aspect.__functionArgs` is for preserving them.

I noticed no test was failing with your `__functionArgs` having `default = { }`, how did you noticed the infinite recursion if no test fails?

vic: > Alright what about this approach of special casing the freeformType

lgtm, we already have special functorType to avoid merging function results which was causing duplicates.

DylanRJohnston: > I noticed no test was failing with your __functionArgs having default = { }, how did you noticed the infinite recursion if no test fails?

It fails on my local machine, unclear why it doesn't fail in CI.

DylanRJohnston: Added a new higher order type 'partitionedType` which partitions the defs and then merges them using different types.

DylanRJohnston: So we can then poke a hole in it to let __functionArgs through unscathed without triggering any of the need for default option definitions.

```nix
freeformType = partitionedType {
  name = "Aspect Freeform Type";
  description = "Aspect freeform type, has a hole in it for __functionArgs";
  partition = def: def.value ? __functionArgs;
  baseType = lib.types.lazyAttrsOf lib.types.deferredModule;
};
```

vic: > It fails on my local machine, unclear why it doesn't fail in CI.

Hm, sounds like tests are useless then, otherwise we won't notice if we break things in the future.

DylanRJohnston: We could also put __functor through this as well, and then we can check that __functionArgs doesn't appear without __functor. Right now it doesn't get into this merge function because it's captured by the option definition.

vic: Yes, please include __functor in the partition, and after this, do we need to keep the strip for functor and functionArgs we had at namespace.nix ? I guess we do, right? because we are working at not merging them, not at how they are propagated beyond the eval/flake frontier.

DylanRJohnston: I tried to do this, but I couldn't quite get it to work, I'm ending up with weird partitions like this

```
[
    {
        right = [ ];
        wrong = [
            { 
                file = "<unknown-file>";
                value = { 
                    __functor = { 
                        _type = "override";
                        content = «lambda ctxApply @ /nix/store/abn8hdayw2gg44dnlrzj7zbjg8ydkywq-source/nix/lib/ctx-apply.nix:101:14»;
                        priority = 50;
                    };
                };
            }
        ];
    }
    {
        right = [ 
            { 
                file = "<unknown-file>";
                value = { __functionArgs = { host = false; }; };
            }
            { 
                file = "<unknown-file>";
                value = { 
                    __functor = «lambda __functor @ /nix/store/660ljssgj5ky0ph0z8pwc1jw8q39xb6j-source/lib/types.nix:1113:11»;
                };
            }
        ];
        wrong = [ ];
    }
]
```

DylanRJohnston: I think it comes from `nix/lib/ctx-types.nix`, in `ctxSubmodule` `config.__functor = lib.mkForce ctxApply` which I don't think you're supposed to do, because then __functor isn't a function until module resolution and it gets fully resolved. But my brain hurts from staring at this stuff all day, so I don't feel like I can figure it out properly right now

vic:
Fixed now, removed our old `functorType` and replaced it with a simpler `lastFunctionTo` type, I believe it was the functorType which was wrong, I also removed a lot of dirty cleanup hacks we did at namespace.nix because of this. Your tests about preserving argument info work now.

vic:
Thanks @DylanRJohnston

PR #354, opened 2026-04-01
feat(batteries): mutual-provider allows user-to-users provides

vic:


templates/ci/modules/features/user-host-mutual-config.nix
```diff
@@ -244,5 +244,53 @@
...
+          tux = tux.programs.vim.enable;
+          alice = alice.programs.vim.enable;
+          bob = bob.programs.vim.enable;
+          carl = carl.programs.vim.enable;
+          aliceTmux = alice.programs.tmux.enable;
+          bobTmux = bob.programs.tmux.enable;
+        };
+
+        expected = {
+          tux = false;
```
drupol, line 284: I don't really understand why this is false here.

vic: we avoid user providing to itself via mutual-provider. see filter at implementation

vic: for all users we have host._.to-users

drupol: Thank you :)

PR #366, opened 2026-04-03
docs: add Flake Parts Modules to documentation sidebar

theutz:


vic:
Thanks a lot, @theutz

PR #368, opened 2026-04-03
fix: ensure example builds on non-linux machines

theutz:
Since the only host in the example didn't match my architecture, `nix fmt` wouldn't run properly. This fixed it.

### DIFF
--- a/templates/flake-parts-modules/modules/custom-classes.nix
+++ b/templates/flake-parts-modules/modules/custom-classes.nix
@@ -6,7 +6,7 @@
   ...
 }:
 {
-  systems = builtins.attrNames den.hosts;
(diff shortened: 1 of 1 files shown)


vic:
Keep the local change on your end, this was intended to show that systems are not hardcoded nor from `nix-systems/default` input, but those used by hosts.

theutz:
Ah, nice. Yeah, I actually stole the idea and implemented it in my own flake, too. :D Do you think it'd be worth putting a comment in-place there to explain that for us Darwin users? Or do you think it's enough to have this issue in place so folks can search for it?

vic:
@theutz yeah, a comment about assigning systems from den hosts might be a better change.

PR #370, opened 2026-04-03
fix: don't make perSystemFwd a function

theutz:
Fixes #367


theutz:
Fixes #371

vic:
I guess we need a an aditional flake check at that template because that bug was not catched by any of the CI checks. Feel like adding some check that help us prevent it again ? 

EDIT: I will merge, but if you want to send another PR with a flake check for this feel free.

theutz:
Will do! Interestingly, the checks passed... but only because they effectively didn't run. :)

PR #397, opened 2026-04-08
feat: per-aspect adapter accumulation in resolve

sini:
Aspects can declare meta.adapter (typed nullOr functionTo raw) to compose into resolution for their subtree. resolve's go function accumulates aspect-level adapters as it descends, composing them with the inherited adapter.

### DIFF
--- a/nix/lib/aspects/resolve.nix
+++ b/nix/lib/aspects/resolve.nix
@@ -31,10 +31,17 @@ let
           aspect-chain = prevChain ++ [ provided ] ++ (lib.optional (provided != aspect) aspect);
 
           classModule = lib.optional (aspect ? ${class}) (
-            lib.setDefaultModuleLocation "${class}@${aspect.name}" aspect.${class}
+            lib.setDefaultModuleLocation "${class}@${aspect.name or "<anon>"}" aspect.${class}
           );
 
           recurse = go aspect-chain;
+
+          resolveChild =
+            i:
+            apply i {
+              inherit class;
+              aspect-chain = aspect-chain ++ [ i ];
+            };
         in
         adapter {
           inherit
@@ -43,12 +50,13 @@ let
             classModule
             recurse
             aspect-chain
+            resolveChild
             ;
         };
     in
     go [ ];
 
-  resolve = withAdapter adapters.module;
+  resolve = withAdapter (adapters.filterIncludes adapters.module);
 
 in
 {
--- a/nix/lib/aspects/types.nix
+++ b/nix/lib/aspects/types.nix
@@ -101,7 +101,12 @@ let
             description = "Aspect attached meta data";
             type = lib.types.submodule {
               freeformType = lib.types.lazyAttrsOf lib.types.unspecified;
-              self = config;
+              config.self = config;
+              options.adapter = lib.mkOption {
+                description = "Adapter to compose into resolution for this aspect's subtree";
+                type = lib.types.nullOr (lastFunctionTo lib.types.raw);
+                default = null;
+              };
             };
             defaultText = lib.literalExpression "{ }";
             default = { };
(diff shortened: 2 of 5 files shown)


nix/lib/aspects/types.nix
```diff
@@ -101,7 +101,12 @@ let
             description = "Aspect attached meta data";
             type = lib.types.submodule {
               freeformType = lib.types.lazyAttrsOf lib.types.unspecified;
-              self = config;
+              config.self = config;
+              options.adapter = lib.mkOption {
+                description = "Adapter to compose into resolution for this aspect's subtree";
+                type = lib.types.nullOr (lib.types.functionTo lib.types.raw);
```
vic: Please change this `lib.types.functionTo` to be our `lastFunctionTo`. We use that because of `functionTo` merge semantics: It allows many different definitions of the same option (different functions) and merges their final result. This was causing many duplicate errors at Den when we had functionTo as type for our __functor. So just to prevent same kind of duplicates, use lastFunctionTo.

nix/lib/aspects/resolve.nix
```diff
@@ -34,19 +34,44 @@ let
...
 
-          recurse = go aspect-chain;
+          subtreeAdapter =
+            if aspect.meta.adapter or null != null
+            then aspect.meta.adapter currentAdapter
+            else currentAdapter;
+
+          recurse = go subtreeAdapter aspect-chain;
+
+          probeInclude =
```
vic: I would love if you (or your agent) can move all this logic of `probeInclude` and `currentAdapter` and `subtreeAdapter` into an adapter that composes with `adapters.module`, and use that composite adapter as the default for our `resolve = withAdapter <yourComposedAdapterHere>`

Maybe it is possible to have no logic changes on this file, but only change the adapter being used by default at `resolve = `.

vic (changes requested):

nix/lib/aspects/types.nix
sini: Easy enough, consider it done.

nix/lib/aspects/resolve.nix
sini: I'll give that a shot. :+1: I was trying to minimize the number of files changed/touchpoints, but thats not a bad place for that logic. Thanks for the feedback!

nix/lib/aspects/resolve.nix
```diff
@@ -24,7 +24,7 @@ let
     adapter: class:
     let
       go =
-        prevChain: provided:
+        currentAdapter: prevChain: provided:
```
vic: Hm, I was actually trying you to move this `currentAdapter` state out of this file, and make it part of another function at adapters.nix, I'm trying to keep resolve.nix generic enough, without having filtering includes in this file. Filtering includes is part of the "per-adapter accumulation" adapter. 

Try moving all state-handling and filtering logic out of this file.

vic (changes requested):

sini:
For the remaining resolve changes:
 ▎ resolveChild 
 This exposes apply to adapters so they can evaluate raw includes (which may be functions/functors). The filterIncludes  adapter needs this in order to probe each include against the composed adapter before deciding whether to keep it — recurse isn't suitable since it performs a full resolve and never returns {}.                                      
                                                                                                                                                                                                
Resolve default applies the filterIncludes module because module alone doesn't act on meta.adapter, so filterIncludes is what reads it, composes the filter, probes includes, and tags survivors for propagation.

If you want I can merge it into module, or keep them as separate compositions.

vic (dismissed):
Awesome! looks good.

vic:
Run `just fmt`

sini:
All clean now.

vic:
Awesome work @sini, you rock! :100:

PR #398, opened 2026-04-08
feat: preserve aspect identity and adapter through functor evaluation

sini:
Functor-producing functions in parametric.nix (applyIncludes, deepRecurse, withOwn) carry name, __provider, and eta.adapter from self into their return values. ctxApply also carries name and __provider on its result.

Currently, when parametric functors evaluate, they return plain { includes = [...]; } attrsets — losing the aspect's identity. This means any code inspecting the resolved aspect (such as a custom adapter via resolve.withAdapter) sees incorrect defaults instead of the original aspect's name.

This lays groundwork for context-level adapter resolution, where adapters declared on aspects or contexts need to propagate through the parametric pipeline to reach forward.nix and outputs.nix. Child includes are tagged with the parent's meta.adapter so adapters compose downward through the include tree.

Upcoming related PRs:
- #397
- Structural provider provenance via __provider paths
- Context-level adapter on ctxApply result
- Forward/output consumption of meta.adapter

### DIFF
--- a/nix/lib/ctx-apply.nix
+++ b/nix/lib/ctx-apply.nix
@@ -105,14 +105,16 @@ let
       result = [ ];
     } items).result;
 
-  ctxApply = self: ctx: {
-    includes = assembleIncludes (traverse {
-      prev = null;
-      prevCtx = null;
-      key = self.name;
-      inherit self ctx;
-    });
-  };
+  ctxApply =
+    self: ctx:
+    parametric.withIdentity self {
+      includes = assembleIncludes (traverse {
+        prev = null;
+        prevCtx = null;
+        key = self.name;
+        inherit self ctx;
+      });
+    };
 
 in
 ctxApply
--- a/nix/lib/parametric.nix
+++ b/nix/lib/parametric.nix
@@ -3,13 +3,32 @@ let
   inherit (den.lib) take;
   inherit (den.lib.statics) owned statics isCtxStatic;
 
+  # Preserve aspect identity through functor evaluation.
+  # Carries name and typed meta options so adapters and provenance
+  # survive, without leaking freeform user meta to child results.
+  withIdentity =
+    self: extra:
+    let
+      meta = self.meta or { };
+    in
+    {
+      name = self.name or "<anon>";
+      meta = {
+        adapter = meta.adapter or null;
+        provider = meta.provider or [ ];
+      };
+    }
+    // extra;
+
   parametric.applyIncludes =
     takeFn: aspect:
     aspect
     // {
-      __functor = self: ctx: {
-        includes = (builtins.filter (x: x != { })) (map (fn: takeFn fn ctx) (self.includes or [ ]));
-      };
+      __functor =
+        self: ctx:
+        withIdentity self {
+          includes = builtins.filter (x: x != { }) (map (fn: takeFn fn ctx) (self.includes or [ ]));
+        };
     };
 
   mapIncludes =
@@ -35,7 +54,7 @@ let
       __functor =
         self:
         { class, aspect-chain }:
-        {
+        withIdentity self {
           includes = [
             (include self { inherit class aspect-chain; })
             (mapIncludes (deepRecurse include branch leaf) leaf (branch aspect))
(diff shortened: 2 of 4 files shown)


nix/lib/parametric.nix
```diff
@@ -3,13 +3,23 @@ let
...
   parametric.applyIncludes =
     takeFn: aspect:
     aspect
     // {
-      __functor = self: ctx: {
-        includes = (builtins.filter (x: x != { })) (map (fn: takeFn fn ctx) (self.includes or [ ]));
-      };
+      __functor = self: ctx:
+        let adapter = self.meta.adapter or null;
+        in {
```
vic: Create a function that aguments a given aspect with this:

```nix
aspect // {
  name = aspect.name;
  meta = { inherit adapter; };
  includes = map (tagAdapter adapter) aspect.includes;
}
```

and use that function

nix/lib/parametric.nix
```diff
@@ -3,13 +3,23 @@ let
...
+          name = self.name;
+          __provider = self.__provider or [];
```
vic: This patch only reads `self.__provider` to preserve it, but it has no definitions of what `self.__provider` is other than empty.

Seems like `__provider` is some kind of metadata, it would be much better if we simply preserve `meta`. Assuming it is better for `__provider` to be `meta.provider`, since I don't know from this patch what `__provider` is.

Does preserving the whole `.meta` makes sense to you ?

vic (dismissed):

nix/lib/parametric.nix
sini: I was torn because name is first class, and this was as important as .name in my head. However, it /can/ live in meta -- nothing is stopping it. I'll update the PR to move it there -- we can always promote it to the core aspect sub-module definition later.

nix/lib/ctx-apply.nix
```diff
@@ -105,7 +105,14 @@ let
       result = [ ];
     } items).result;
 
-  ctxApply = self: ctx: {
+  ctxApply = self: ctx:
```
sini: I'll update this to pull the same lib def of withIdentity.

sini:
Alright, updated.

nix/lib/parametric.nix
```diff
@@ -3,12 +3,25 @@ let
...
   inherit (den.lib.statics) owned statics isCtxStatic;
 
+  # Preserve aspect identity through functor evaluation.
+  # Carries name and typed meta options so adapters and provenance
+  # survive, without leaking freeform user meta to child results.
+  withIdentity = self: extra:
+    let meta = self.meta or {};
+    in {
+      inherit (self) name;
+      meta = {
```
vic, line 16: please preserve also the metadata added by Den, `meta.loc or [ ]`, `meta.name or self.name` (full aspect name based on loc), `meta.file or "<unknown>"`.

nix/lib/parametric.nix
```diff
@@ -64,7 +77,7 @@ let
     functor: aspect:
     aspect
     // {
-      __functor = self: ctx: {
+      __functor = self: ctx: withIdentity self {
```
vic: much better :) I like this `withIdentity` combinator.

nix/lib/parametric.nix
sini: Alright, I'll do a more complicated scan of existing meta keys now that it's defined in a single place. 👍

vic: from templates/ci/modules/features/aspect-meta.nix

`expected.KEYS = "file:foo:loc:name:self";` (except for `foo` all others are Den metadata), not sure we want to preserve `self` it is a reference to the aspect itself.

sini: I was about to just implement your comment but upon reflection -- I'm currently under the impression the attrset keys for meta should still merge from the submodule default definition, and I shouldn't need to explicitly copy them. Am I wrong in this? I'll do some testing to verify.

sini: Alright, I've got a test working and it provides values like `childFile = "<curried>"` so maybe it's best to forward overwrite them in this context. Adding it now.

sini: `meta.loc`, `meta.name`, and `meta.file` are set by aspectMeta via mkForce at merge time and resolve to empty/generic defaults ("", [], "<curried>") during functor evaluation. There's nothing meaningful to preserve so withIdentity can't copy values that which don't exist yet. The curried path through aspectType.merge already provides equivalent defaults.

sini: pushing a test to exercise this anyways

vic: right, forgot that those don't exist yet.

sini:
awesome. I have 3 more commits staged but I'll save them for tomorrow. Goodnight and thank you for the reviews!

edit: (what do you mean self.name isn't set... okay this is an easy fix... pushing in a minute.)

vic:
Awesome, run `just fmt` and we are done with this. Looking forward to your other PRs.

Good job so far, @sini. Thank you so much.

PR #399, opened 2026-04-08
feat: structural provider provenance via meta.provider

sini:
**Summary**
- Adds meta.provider typed option tracking each aspect's structural origin as a path (e.g., ["foo" "bar"] for foo.provides.bar)
- providerPrefix threads through the provides submodule so child providers accumulate their parent's path
- mkAspectsType factory allows namespaces and den.provides to set their own provider root

### How it works

The provides submodule extends providerPrefix with the current aspect's name before passing it to child provider types. Namespaces set providerPrefix = [ name ] as their root. den.provides uses [ "den" "provides" ]. Top-level aspects default to [].

This lets adapters distinguish "bar provided by foo" from a top-level "bar" by inspecting aspect.meta.provider.

### DIFF
--- a/modules/aspects/provides.nix
+++ b/modules/aspects/provides.nix
@@ -10,7 +10,14 @@
         default = { };
         description = "Batteries Included - re-usable high-level aspects";
         type = lib.types.submodule {
-          freeformType = lib.types.attrsOf config.den.lib.aspects.types.providerType;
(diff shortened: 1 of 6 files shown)


vic:
Flawless!

Please merge this yourself when you accept the maintainer invitation.

PR #405, opened 2026-04-08
feat: forward and output resolution respects meta.adapter

sini:
## Summary
- Entities automatically expose `.resolved` — the result of their context pipeline — derived from the schema type system
- `forward.nix` defaults to `item.resolved` when `fromAspect` is absent, enabling cross-context forwarding without manual wiring
- `mainModule` helper eliminated — entity types resolve directly via `config.resolved`

## What changed
**options.nix** — `schemaEntryType` wraps `deferredModule` to auto-inject `config.resolved` for any schema entry where `den.ctx.${kind}` exists. Context args are derived from the entity's `_module.args`, filtered to known context kinds. No per-entity boilerplate — host, user, and home all get `.resolved` automatically.

**types.nix** — `mainModule` helper removed. Both host and home `mainModule` options simplified to `den.lib.aspects.resolve config.class config.resolved`.

**forward.nix** — `asp` falls back to `item.resolved or item` when `fromAspect` is absent.

## How context adapters flow through forwards
`den.ctx.host.meta.adapter` is carried through `ctxApply` (via `withIdentity` from #398) onto `.resolved`. When `resolve` processes it, `adapters.filterIncludes` (from #397) picks up the adapter and applies it transitively to the entire subtree — including nested aspects reached through forwards.

### DIFF
--- a/modules/options.nix
+++ b/modules/options.nix
@@ -14,12 +14,51 @@ let
       config
       ;
   };
+
+  # Schema entries auto-inject config.resolved when den.ctx.${kind} exists.
+  # Context args are derived from the entity's _module.args, filtered to
+  # known context kinds so framework args don't leak through.
+  schemaEntryType =
+    let
+      base = lib.types.deferredModule;
+    in
+    base
+    // {
+      merge =
+        loc: defs:
+        let
+          kind = lib.last loc;
+          merged = base.merge loc defs;
+          resolvedCtx =
+            { config, ... }:
+            {
+              options.resolved = lib.mkOption {
+                description = "The resolved aspect for this ${kind}, produced by den.ctx.${kind}.";
+                readOnly = true;
+                type = lib.types.raw;
+                default = den.ctx.${kind} (
+                  lib.filterAttrs (n: _: den.ctx ? ${n}) config._module.args // { ${kind} = config; }
+                );
+              };
+            };
+        in
+        if den.ctx ? ${kind} then
+          {
+            imports = [
+              merged
+              resolvedCtx
+            ];
+          }
+        else
+          merged;
+    };
+
   schemaOption = lib.mkOption {
     description = "freeform deferred modules per entity kind";
     defaultText = lib.literalExpression "{ }";
     default = { };
     type = lib.types.submodule {
-      freeformType = lib.types.lazyAttrsOf lib.types.deferredModule;
+      freeformType = lib.types.lazyAttrsOf schemaEntryType;
     };
   };
 in
--- a/nix/lib/types.nix
+++ b/nix/lib/types.nix
@@ -103,8 +103,8 @@ let
             visible = false;
             readOnly = true;
             type = lib.types.deferredModule;
-            defaultText = ''den.lib.aspects.resolve "nixos" (den.ctx.host { inherit host; })'';
-            default = mainModule config den.ctx.host "host";
+            defaultText = "den.lib.aspects.resolve config.class config.resolved";
+            default = den.lib.aspects.resolve config.class config.resolved;
           };
         };
       }
@@ -258,19 +258,12 @@ let
             visible = false;
             readOnly = true;
             type = lib.types.deferredModule;
-            defaultText = lib.literalExpression "mainModule";
-            default = mainModule config den.ctx.home "home";
+            defaultText = "den.lib.aspects.resolve config.class config.resolved";
+            default = den.lib.aspects.resolve config.class config.resolved;
           };
         };
       }
     );
-
-  mainModule =
-    from: intent: name:
-    let
-      asp = intent { ${name} = from; };
-    in
-    den.lib.aspects.resolve (from.class) asp;
 in
 {
   inherit hostsOption homesOption;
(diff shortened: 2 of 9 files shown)


modules/outputs.nix
```diff
@@ -6,7 +6,8 @@
 }:
 let
   no-flake-parts = !inputs ? flake-parts;
-  flakeModule = den.lib.aspects.resolve "flake" (den.ctx.flake { });
+  flakeCtx = den.ctx.flake { };
```
vic: The result of applying a ctx is an aspect, since that is the result of all ctx transformations, it makes more sense to name this `flakeAspect`, let's use the `ctx` terminology for things that can transform via `.into`.

templates/ci/modules/features/cross-context-forward.nix
```diff
@@ -0,0 +1,261 @@
...
+
+        den.aspects.iceberg.includes = [
+          (
+            { host }:
+            den._.forward {
+              each = lib.attrValues den.hosts.${host.system};
+              fromClass = _: "test-class";
+              intoClass = _: host.class;
+              intoPath = _: [ ];
+              sourceGuard = src: src != host;
```
vic: I'm not sure about `sourceGuard` this is the only place I see it being set, a better name would be `itemGuard`, since we use the `item` term to refer to each element of the `each` list.

Most importantly: I don't see the point of having `sourceGuard` when you can simply say `each = filter (i: i != host) hosts`

nix/lib/types.nix
vic: remove `mainModule =`, now that each entity has its own `.ctx` we can instead define at entity types:

```nix
mainModule = lib.mkOption {
  default =
    den.lib.aspects.resolve config.class config.ctx;
};
```

nix/lib/types.nix
```diff
@@ -28,6 +28,7 @@ let
         freeformType = lib.types.attrsOf lib.types.anything;
         imports = [ den.schema.host ];
         config._module.args.host = config;
+        config.ctx = den.ctx.host { host = config; };
```
vic: I like this, but I'm not sure about the `.ctx` name, because in my mind a context is a pipeline of transformations, NOT the result of applying such transformations, which is an aspect.

vic (changes requested):

nix/lib/types.nix
sini: Alright, hold on for a pending refactor. I'm going to try and push this up the stack a bit and instead treat host-user as a special over-ride for a more general `${kind} = self` entityCtx composition

sini: Still in final drafting phase -- sorry for opening this a little prematurely. :)

templates/ci/modules/features/cross-context-forward.nix
sini: Yeah, there's no point. It's a slip-in from another feature for globalized cross-context forwarding that's not ready that I thought needed source and destination guards, and I may never submit. :) Removed here.

nix/lib/types.nix
sini: It's gone! Good call.

modules/outputs.nix
sini: Folded in, it doesn't need to be named at all.

nix/lib/types.nix
sini: Okay, /now/ it's review ready but your comments were helpful!

nix/lib/types.nix
```diff
@@ -137,6 +143,13 @@ let
       }
     );
 
+  entityCtx =
```
sini: I'm going to explore if I can fold this into schema.

sini: And I can. Done.

sini: Maybe I can clean the interface up even more. Working on that.

modules/options.nix
```diff
@@ -22,15 +22,39 @@ let
       freeformType = lib.types.lazyAttrsOf lib.types.deferredModule;
     };
   };
+
+  # Default resolved context for an entity kind: den.ctx.${kind} { ${kind} = config; }
+  entityCtx =
+    kind:
+    { config, ... }:
+    {
+      config.resolved = lib.mkDefault (den.ctx.${kind} { ${kind} = config; });
```
vic: I think the right place for this was at types.nix, since there's `defaultModule` which depends on `resolved`.

Also, it might be better to define a proper option, so we can attach description documentation to it. (typed:  `raw` to prevent it from having merge-semantics).

```nix
options.resolved = lib.mkOption {
  description = ''
   The resolved aspect for this ${kind}.

   If you want to use a custom ctx pipeline
  other than the provided by Den,
  override this value.
  '';
  default = den.ctx.${kind} { ... };
};
```

vic: Looking good, we might need an small update to the schema docs, where we list the options each entity has, to mention that people can set this `resolved` to use custom pipelines.

vic (changes requested):

modules/options.nix
sini: Yeah, moved it back to draft while I try to simplify the interface a bit more.

I'll add docs to the scope of the final draft.

vic (approved):

vic:
Amazing, @sini! keep those stacked PRs coming.

PR #406, opened 2026-04-08
BUG: No merge semantics on den aspects defined multiple times

vic:
From https://matrix.to/#/!mafjNSATIwbTXcmNzX:matrix.org/$XldtkLSIBLeUFAQKLP7jHQNHOx6HVbMV95T2Xe6kUpc?via=matrix.org&via=rattlesnake.fr&via=vates.tech

### DIFF
--- a/templates/bogus/modules/bug.nix
+++ b/templates/bogus/modules/bug.nix
@@ -15,10 +15,12 @@
         den.hosts.x86_64-linux.igloo.users.tux = { };
 
         den.aspects.foo = { host }: {
-          nixos = lib.optionalAttrs (host.name == "igloo") {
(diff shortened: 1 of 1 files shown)


templates/bogus/modules/bug.nix
```diff
@@ -14,11 +14,37 @@
         # replace <system> if you are reporting a bug in MacOS
         den.hosts.x86_64-linux.igloo.users.tux = { };
 
-        # do something for testing
-        den.aspects.tux.user.description = "The Penguin";
+        imports = [
```
vic, line 17: Currently den aspects have no merge semantics, we use only one function definition unlike `functionTo`.

vic:
Fixed by #410

Issue #246, opened 2026-03-09
CI: detect bogus dirty-test to quickly test something without running all CI


vic:

vic:
done

PR #408, opened 2026-04-08
BUG: aspect being called as if it was a module-function

vic:
EDIT: we need a new CI action that runs bogus when it detect we change a .nix file on templates/bogus. Instead of having to approve this for templates to run at CI. #246 

Fails with

```console
       error: attribute 'host' missing
       at /nix/store/2n2i983z59kjmw707xpddjdywyk8vn2g-source/lib/modules.nix:726:13:
          725|             "noting that argument `${name}` is not externally provided, so querying `_module.args` instead, requiring `config`"
          726|             config._module.args.${name}
             |             ^
          727|           )
```

### DIFF
--- a/templates/bogus/modules/bug.nix
+++ b/templates/bogus/modules/bug.nix
@@ -14,11 +14,30 @@
         # replace <system> if you are reporting a bug in MacOS
         den.hosts.x86_64-linux.igloo.users.tux = { };
 
-        # do something for testing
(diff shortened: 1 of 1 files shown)


sini:
Looking into this, I'm reproducing this on my end as well.

sini:
I have the root cause and a fix, working on the cleanest possible version of the fix.

PR #409, opened 2026-04-08
feat: tombstones, identity paths, and substitution in filterIncludes

sini:
filterIncludes now produces tombstones (~name, meta.excluded) for excluded includes instead of silently dropping them. Tombstones are empty aspects harmless to module but visible to trace/debug adapters.

Substituted includes produce a tombstone for the original plus the replacement, enabling both to appear in traces.

New adapters:
- aspectPath: derive identity from name + provider (replaces == on aspects)
- excludeAspect: exclude by aspect reference via path comparison
- substituteAspect: substitute by aspect reference via mapAspect + path
- tombstone: create a tombstone from a resolved aspect (exported utility)

### DIFF
--- a/nix/lib/aspects/adapters.nix
+++ b/nix/lib/aspects/adapters.nix
@@ -1,16 +1,12 @@
-# Adapters for resolve.withAdapter. Default one is module.
+# Adapters for resolve.withAdapter. Default adapter is module.
 #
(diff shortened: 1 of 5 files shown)


vic (approved):
lgtm, merge when you feel like it.

PR #410, opened 2026-04-08
fix: perHost/perUser/perHome preserve aspect identity

sini:
Return { includes = [fn] } instead of a bare function, matching the POC pattern. The function result lives in includes where parametric.applyIncludes handles it with withIdentity, preserving the aspect name through resolution. Fixes #408.

### DIFF
--- a/modules/context/perHost-perUser.nix
+++ b/modules/context/perHost-perUser.nix
@@ -7,9 +7,9 @@
 let
   inherit (config.den.lib) take parametric;
   fixed = ctx: aspect: parametric.fixedTo ctx { includes = [ aspect ]; };
-  perHost = aspect: take.exactly ({ host }@ctx: fixed ctx aspect);
-  perUser = aspect: take.exactly ({ host, user }@ctx: fixed ctx aspect);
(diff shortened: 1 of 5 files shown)


vic (dismissed):
:O

sini:
Don't merge yet, it fixed the test but theres one other edgecase I want to check.

vic:
merge at your discretion

vic:
please also include a deadbug test with the original code.

sini:
So this was an interaction with:
```nix
  providerFnType =
    cnf:
    let
      eth = lib.types.either (leafProviderFnType cnf) (curriedProviderFnType cnf);
    in
    eth
    // { # <-- This bit here was the culprit, the updated signatures allow us to keep it and should work now
      merge =
        loc: defs:
        (aspectType cnf).merge loc [
          {
            file = (lib.last defs).file;
            value = {
              __functor = _: eth.merge loc defs;
            };
          }
        ];
    };
    ```

vic (approved):

vic:
Thanks, @sini

PR #414, opened 2026-04-09
feat: withSubAspectsOf

theutz:
Add a `withSubAspectsOf` helper to the den lib to be used like so:

```nix
{ den, ... }:
den.aspect.foo = {
  includes = den.lib.withSubAspectsOf [ "foo" ] [ den.aspects.bar ];
}
```

The function maps over the first list, grabbing any `provides` from the aspects enumerated therein. The second list is concatenated with the first, and functions like the normal `includes`.

Thought this might be a less magical way to reduce boilerplate than #396, while still providing a nice bit of DX.


theutz:
@vic I started using this in my personal config and it works fine, but there are two test cases that aren't passing. I think it might be because I'm not setting up the "mock" values in the tests properly?

theutz:
Also, as usual, I'll leave it to you to decide whether this is needed or not!

theutz:
> @vic I started using this in my personal config and it works fine, but there are two test cases that aren't passing. I think it might be because I'm not setting up the "mock" values in the tests properly?

Oh, dear. Weirdly, the last two tests don't pass for me on my local, but they pass here...

templates/ci/modules/features/with-sub-aspects-of.nix
```diff
@@ -0,0 +1,56 @@
...
+      { den, ... }:
+      {
+        den.aspects.foo = _: { };
+
+        expr = den.lib.withSubAspectsOf [ ] [ den.aspects.foo ];
+        expected = [ den.aspects.foo ];
+      }
+    );
+
+    test-works-with-sub-aspects = denTest (
```
theutz, line 30: When I run `just ci` locally, this test and the following one fail.

nix/lib/with-sub-aspects-of.nix
```diff
@@ -0,0 +1,18 @@
...
+  withSubAspectsOf =
+    aspects: includes:
+    lib.addErrorContext "while collecting sub aspects for ${lib.join "," aspects}" (
+      assert lib.isList aspects;
+      assert lib.isList includes;
+      (lib.map (
+        aspect:
+        lib.addErrorContext "while collecting sub aspect for ${aspect}" (
+          assert lib.isString aspect;
+          lib.getAttrFromPath [ "aspects" aspect "provides" ] den
```
theutz, line 12: If there is, in fact, something wrong with the implementation, I wonder if it's that I'm using the `den` value from the outer closure. That didn't feel quite right.

vic:
> Add a `withSubAspectsOf` helper to the den lib to be used like so:
> 
> ```nix
> { den, ... }:
> den.aspect.foo = {
>   includes = den.lib.withSubAspectsOf [ "foo" ] [ den.aspects.bar ];
> }
> ```
> 
> The function maps over the first list, grabbing any `provides` from the aspects enumerated therein. The second list is concatenated with the first, and functions like the normal `includes`.
> 
> Thought this might be a less magical way to reduce boilerplate than #396, while still providing a nice bit of DX.

I have a few issues with this:

I don't like using string-typed references, like "foo" is that a value or represents den.aspect.foo ? why not use den.aspect.foo reference directly. Many nix frameworks
use string-typed references I don't want that.

There's no need for the second list if that is just appended, let people do that. 

What about:

```nix
providesOf = aspect: { includes = lib.attrValues aspect.provides; };
```

The only difference between `providedOf` and using `attrValues` directly is that
the first creates an aspect (keeps things in the same aspect domain of den).

Then you can use it also in other places


```nix
den.aspects.foo.includes = 
 (map providesOf [ den.aspects.foo den.aspects.bar ]) ++ 
 [ den.aspects.other ]
```

or another recursive variant:

```nix
deepProvidesOf = aspect: {
  includes = map (a: [a] ++ (deepProvidesOf a).includes) aspect.provides;
}
```

but then they look very similarly to how adapters are defined. (see adapters.module for example)

theutz:
> I don't like using string-typed references, like "foo" is that a value or represents den.aspect.foo

Yeah, that makes sense to me.

> Many nix frameworks use string-typed references I don't want that.

Good point. I'm on board!

> `providesOf = aspect: { includes = lib.attrValues aspect.provides; };`

I like the shape and signature of this, yeah. I would maybe suggest changing the name to something like `providedBy` or `aspectsProvidedBy`.

The reason I think a feature like this is important is not because people shouldn't be forced to write `lib.attrValues`. I think it's a way for us to use the API to guide new users to the most common tasks. It puts a semantic wrapper around the operation.

> but then they look very similarly to how adapters are defined. (see adapters.module for example)

But, as with the glob stuff, I'll be happy to defer writing any kind of convenience methods, syntax sugar, or helper APIs until the underlying plumbing is in place. :D

theutz:
Thanks for letting me use these PRs to increase my knowledge of Nix and Den! I hit the `approved` label so the tests would run so you could see what I was doing wrong. The more I learn about how to test these things, the better my future PRs will be! ☘️

vic:
So the test should be failing ? they are green. 

Haven't looked at what they do.

theutz:
Well, I don't know if I can say authoritatively if the should be failing. But they were failing when I ran `just ci` locally. If it's the "Check Template" matrix test that's supposed to be running it, it looks like those got skipped again for some reason.

vic:
`providesOf` of whatever name you choose (I'm bad at naming, you know better english), reminds me a bit of the combinators we have at `parametric`, like you say, `parametric` is a helper for setting `__functor`, we also have some other operations on aspects that return eg, only the owned attribute (removes includes and __functor), or only static includes, etc.  So, something like `providesOf <aspect>` returning another aspect makes sense to me.

theutz:
I'll go ahead and close this since we aren't gonna go forward with this specific helper currently, and I'll try and make a bogus test PR to see if I can reproduce the error just for learning purposes. 👨‍🎓

PR #419, opened 2026-04-09
fix: provider sub-aspect functions receive parametric context

sini:
Fixes #413

### Problem

Provider sub-aspects defined as bare context functions don't receive `{ host }` during resolution:

```nix
den.aspects.foo._.sub = { host, ... }: {
  nixos = lib.optionalAttrs (host.hostName != "whatever") {
    networking.networkmanager.enable = true;
  };
};
```

When a parent function returns `{ includes = [foo._.sub] }`, the sub-aspect is resolved by the adapter system which only passes `{ class, aspect-chain }` — the `{ host }` context from the pipeline never reaches nested includes of function results.

### Fix

`applyDeep` in `parametric.applyIncludes` — when `takeFn` succeeds and returns a bare result with sub-includes (no `meta`, no `__functor`), also apply `takeFn` to those sub-includes. This propagates context to provider sub-aspects nested inside function results without double-applying to parametric wrappers or `deepRecurse` outputs.

The key discriminator: bare provider results carry only `includes` (+ `name` from `carryAttrs`). Results from `withOwn`/`withIdentity` have `meta`; deferred `deepRecurse` wrappers have `__functor`. Neither should be re-resolved.

### Test coverage

- `deadbugs/issue-413-provider-bare-function.nix` — reproduces the original report
- `deadbugs/issue-413-provider-sub-aspect-function.nix` — variant with `lib.optionalAttrs` guard
- `provides-parametric.nix` — provider sub-aspects with parametric context in various configurations

cc: @kalyanoliveira for the report

### DIFF
--- a/nix/lib/parametric.nix
+++ b/nix/lib/parametric.nix
@@ -20,6 +20,37 @@ let
     }
     // extra;
 
+  # When takeFn succeeds and returns a result with sub-includes,
+  # also try to resolve those sub-includes with takeFn. This handles
+  # provider sub-aspect functions nested inside include results:
+  # e.g. wrapped_fn returns { includes = [foo._.sub]; } where foo._.sub
+  # needs parametric context applied before reaching the static pipeline.
+  applyDeep =
+    takeFn: ctx: fn:
+    let
+      r = takeFn fn ctx;
+      # Bare provider results carry only includes (+ name from carryAttrs).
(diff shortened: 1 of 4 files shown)


vic (approved):

vic:
Thanks @kalyanoliveira and @sini

PR #420, opened 2026-04-09
feat: tombstones record excludedBy for provenance tracking

sini:
Tombstones now carry meta.excludedBy — the name of the aspect whose meta.adapter caused the exclusion. Combined with the existing meta.provider (structural origin from provides chain), tombstones clearly document both who defines an aspect and who excluded it.

Tombstone meta fields:
- provider    — who defines this aspect (provides chain)
- excludedBy  — who excluded it (adapter source)
- originalName — display name before ~prefix
- replacedBy  — replacement name (substitutions only)

### DIFF
--- a/nix/lib/aspects/adapters.nix
+++ b/nix/lib/aspects/adapters.nix
@@ -66,6 +66,12 @@ let
   # Empty aspect marking an excluded include. ~prefix prevents accidental
   # name collisions with live aspects. Harmless to module, visible to trace.
   # Consumers should check meta.excluded before accessing other aspect fields.
+  # Tombstone fields:
+  #   meta.excluded     — true
+  #   meta.originalName — display name before ~prefix
+  #   meta.provider     — who defines this aspect (structural origin, from provides chain)
+  #   meta.excludedBy   — who excluded it (the aspect whose meta.adapter caused exclusion)
+  #   meta.replacedBy   — name of the replacement (for substitutions only)
   tombstone =
     resolved: extra:
     let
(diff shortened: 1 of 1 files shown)


nix/lib/aspects/adapters.nix
```diff
@@ -66,6 +66,12 @@ let
   # Empty aspect marking an excluded include. ~prefix prevents accidental
   # name collisions with live aspects. Harmless to module, visible to trace.
   # Consumers should check meta.excluded before accessing other aspect fields.
+  # Tombstone fields:
+  #   meta.excluded     — true
```
vic, line 70: do we need meta.excluded ? given it is not present when false, we could be testing for `meta ? excludedBy`

vic: oh but it could be also replacedBy

vic: so is `excluded = true` when either replaced or excluded ?

sini: It could also be used outside of our specific chain.

vic (approved):

nix/lib/aspects/adapters.nix
sini: While it could be derived by either field, it does simplify validation when graph generating.

sini: One is a type tag, the other id debug/trace detail.

excludedBy is probably the wrong name -- I'll update this to excludedFrom, since it's where the tombstone is included

sini: Oh -- you already merged. I'll sneak it into my demo trace -> mermaid graph demo template since it has no consumers, unless you want a micro PR. :)

PR #421, opened 2026-04-09
fix: rename excludedBy to excludedFrom on tombstones

sini:
The value is the parent aspect whose meta.adapter caused the exclusion, not the agent doing the excluding. excludedFrom better describes "this aspect was excluded from server's subtree."

### DIFF
--- a/nix/lib/aspects/adapters.nix
+++ b/nix/lib/aspects/adapters.nix
@@ -70,7 +70,7 @@ let
   #   meta.excluded     — true
   #   meta.originalName — display name before ~prefix
   #   meta.provider     — who defines this aspect (structural origin, from provides chain)
-  #   meta.excludedBy   — who excluded it (the aspect whose meta.adapter caused exclusion)
+  #   meta.excludedFrom — the parent aspect whose meta.adapter caused the exclusion
   #   meta.replacedBy   — name of the replacement (for substitutions only)
   tombstone =
     resolved: extra:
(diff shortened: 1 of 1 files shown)


nix/lib/aspects/adapters.nix
```diff
@@ -70,7 +70,7 @@ let
   #   meta.excluded     — true
   #   meta.originalName — display name before ~prefix
   #   meta.provider     — who defines this aspect (structural origin, from provides chain)
-  #   meta.excludedBy   — who excluded it (the aspect whose meta.adapter caused exclusion)
+  #   meta.excludedFrom — the parent aspect whose meta.adapter caused the exclusion
   #   meta.replacedBy   — name of the replacement (for substitutions only)
```
vic, line 74: so is replacedWith also a better name ? or is replacedBy a reference of who replaced the aspect ?

sini: replacedBy is the aspect that was substituted

vic: oh right, the value of replacedBy was the original aspect

vic: naming is hard, you are much better than me at this.

sini: ex;
`desktop.includes[]` resolves with an adapter sub(regreet,gdm)
`regreet(excluded) -> gdm(inserted)`
so the tombstone for regreet has `excludedFrom = desktop` `replacedBy = gdm`

vic (approved):

PR #423, opened 2026-04-10
BUG: static sub-aspect nixos config silently dropped when included from parametric parent

kalyanoliveira:
## Reproduction

See `templates/bogus/modules/bug.nix`.

The following setup fails — `igloo.networking.networkmanager.enable` is `false` when it should be `true`:

- Schema defines an enable option for a sub-aspect
- Parent aspect is a parametric function that conditionally includes the sub-aspect based on that option
- Sub-aspect is defined as a **static attrset** with nixos config
- Host has the enable option set to `true`

The includes list is correctly populated (length 1) and `host.role._.sub.enable` is correctly `true`, but the sub-aspect's nixos config never reaches the host.

## Fix

Making the sub-aspect a bare parametric function instead of a static attrset makes it work:

```nix
# fails
den.aspects.role._.sub.nixos.networking.networkmanager.enable = true;

# works
den.aspects.role._.sub = { host, ... }: {
  nixos.networking.networkmanager.enable = true;
};
```

## Expected behavior

Static sub-aspects should have their nixos config applied when included from a parametric parent function, just as parametric sub-aspects do.

### DIFF
--- a/templates/bogus/modules/bug.nix
+++ b/templates/bogus/modules/bug.nix
@@ -1,26 +1,48 @@
-{ denTest, ... }:
+{ denTest, lib, ... }:
 {
   flake.tests.bogus = {
-
     test-something = denTest (
+      { den, igloo, ... }:
       {
-        den,
-        lib,
-        igloo, # igloo = nixosConfigurations.igloo.config
-        tuxHm, # tuxHm = igloo.home-manager.users.tux
-        ...
-      }:
-      {
-        # replace <system> if you are reporting a bug in MacOS
         den.hosts.x86_64-linux.igloo.users.tux = { };
 
-        # do something for testing
-        den.aspects.tux.user.description = "The Penguin";
+        imports =
+          let
+            a = {
+              den.schema.host.options.role._.sub.enable = lib.mkEnableOption "sub";
+            };
 
-        expr = igloo.users.users.tux.description;
-        expected = "The Penguin";
+            b = {
+              den.aspects.role = { host, ... }: {
+                includes = lib.optionals host.role._.sub.enable [
+                  den.aspects.role._.sub
+                ];
+              };
+            };
+
+            c = {
+              den.aspects.role._.sub.nixos.networking.networkmanager.enable = true;
+            };
+            # Change c to the below, and it works.
+            # c = {
+            #   den.aspects.role._.sub = { ... }: {
+            #     nixos.networking.networkmanager.enable = true;
+            #   };
+            # };
+
+            d = {
+              den.hosts.x86_64-linux.igloo.role._.sub.enable = true;
+            };
+
+            e = {
+              den.aspects.igloo.includes = [ den.aspects.role ];
+            };
+          in
+          [ a b c d e ];
+
+        expr = igloo.networking.networkmanager.enable;
+        expected = true;
       }
     );
-
   };
 }
--- a/templates/bogus/modules/bug.nix
+++ b/templates/bogus/modules/bug.nix
@@ -8,13 +8,10 @@
 
         imports =
           let
-            a = {
-              den.schema.host.options.enable = lib.mkEnableOption "sub";
-            };
 
             b = {
               den.aspects.role = { host, ... }: {
-                includes = lib.optionals host.enable [
+                includes = [
                   den.aspects.role._.sub
                 ];
               };
@@ -30,15 +27,11 @@
             #   };
             # };
 
-            d = {
-              den.hosts.x86_64-linux.igloo.enable = true;
-            };
-
             e = {
               den.aspects.igloo.includes = [ den.aspects.role ];
             };
           in
-          [ a b c d e ];
+          [ b c e ];
 
         expr = igloo.networking.networkmanager.enable;
         expected = true;
--- a/templates/bogus/modules/bug.nix
+++ b/templates/bogus/modules/bug.nix
@@ -9,12 +9,12 @@
         imports =
           let
             a = {
-              den.schema.host.options.role._.sub.enable = lib.mkEnableOption "sub";
+              den.schema.host.options.enable = lib.mkEnableOption "sub";
             };
 
             b = {
               den.aspects.role = { host, ... }: {
-                includes = lib.optionals host.role._.sub.enable [
+                includes = lib.optionals host.enable [
                   den.aspects.role._.sub
                 ];
               };
@@ -31,7 +31,7 @@
             # };
 
             d = {
-              den.hosts.x86_64-linux.igloo.role._.sub.enable = true;
+              den.hosts.x86_64-linux.igloo.enable = true;
             };
 
             e = {
--- a/nix/lib/parametric.nix
+++ b/nix/lib/parametric.nix
@@ -34,6 +34,26 @@ let
       # wrappers have __functor. Re-resolving either would double-apply
       # context and duplicate modules.
       isBareResult = builtins.isAttrs r && r ? includes && !(r ? meta) && !(r ? __functor);
+      # Full aspect objects have meta set by aspectMeta.
+      isFullAspect = sub: builtins.isAttrs sub && sub ? meta;
+      # Extract only owned class-module keys from a full aspect,
+      # stripping all aspect infrastructure / metadata keys.
+      # If the result is non-empty the sub has static owned class configs
+      # (e.g. nixos = { ... }) that withOwn would skip in a parametric ctx.
+      # Keys declared as options in aspectSubmodule (types.nix).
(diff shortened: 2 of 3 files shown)


kalyanoliveira:
As always, if someone could help me verify that I didn't have any skill issues and that this is indeed a bug, that would be great.

vic:
A bug indeed,

```nix
              den.aspects.role._.sub.nixos.networking.networkmanager.enable = true;
```

this should work. 

Please send a fix on this same PR,

to run locally use (from den root directory)

```
just bogus
```

charlesfire:
OMG. That's why my docker config wasn't  applied.

vic:
If you can simplify the test more, the better.

charlesfire:
This issue also happens with includes :

```
den.aspects.docker = {
    nixos.virtualisation.docker.enable = true;
};

den.aspects.dev = { user, ... }: {
    includes = [ den.aspects.docker ];
}; # docker is never enabled by the dev aspect.
```

templates/bogus/modules/bug.nix
```diff
@@ -1,26 +1,48 @@
...
-      }:
-      {
-        # replace <system> if you are reporting a bug in MacOS
         den.hosts.x86_64-linux.igloo.users.tux = { };
 
-        # do something for testing
-        den.aspects.tux.user.description = "The Penguin";
+        imports =
+          let
+            a = {
```
vic: remove all the part about the sub.enable conditional, not relevant to the bug

vic (changes requested):
If you send a patch solving this, no need to open other PR, just use this same one and when it works we can move the test to a proper file at deadbugs/ and revert changes at bug.nix.

Thanks for finding this one.

kalyanoliveira:
Will do: currently simplifying bug.nix and trying to work on a patch (sponsored by claude^tm)

vic:
> This issue also happens with includes :

omg thanks for that

vic:
> Will do: currently simplifying bug.nix and trying to work on a patch (sponsored by claude^tm)

send it over :) just try to make the patch be clean, no accidental complexity

charlesfire:
Honestly, I'm happy finding this is a bug because for at least a whole hour, I thought I was deeply misunderstanding something.

kalyanoliveira:
Honestly, I don't like the fix that my latest commit introduced. The tests pass, but still, it feels like it's too much hardcoding.

Here's some more context (ai generated):

The latest commit fixed a bug where static and parametric sub-aspects had their nixos config silently dropped when included from a parametric parent.

The root cause is in `applyDeep` in `parametric.nix`: when a sub-aspect is a full aspect (has `meta`), `withOwn` in a parametric context only returns the functor branch — it skips owned class config entirely (e.g. `nixos = { ... }`). So the config was getting lost before it ever reached the resolver.

The fix detects two cases:

1. **Static sub-aspect** (has owned class keys like `nixos = ...`): `classConfig` extracts those keys explicitly and includes them, then recurses into `sub.includes` so nested parametric includes still receive the context.
2. **Parametric sub-aspect** (bare function coerced to `{ includes = [fn] }`): no owned class keys at the top level, so we let `takeFn` handle it — the functor captures the context and propagates it through includes as before.

The tricky part is `classConfig`. It strips the declared option keys of `aspectSubmodule` (`includes`, `__functor`, `__functionArgs`, `name`, `description`, `meta`, `provides`, `_`), leaving only the freeform class config keys. That list is effectively derived from `types.nix`, but the coupling is implicit — there's no structural enforcement.

The question for you: is that acceptable, or would you prefer to define the infrastructure key list once in `types.nix` and import it into `parametric.nix` to make the relationship explicit? Happy to go either way.

templates/bogus/modules/bug.nix
kalyanoliveira: I believe this to be addressed in [932bd7c](https://github.com/vic/den/pull/423/commits/932bd7c4aa7093cf5f648c1dbe9dd5e6dc80a558)

vic:
@kalyanoliveira please read this: https://den.oeiuwq.com/contributing and copy those terms here on the description.

kalyanoliveira:
@vic sorry, but I cannot promise the following:

> You are responsable for problems/bugs caused by your contribution and will promptly fix them.

I do not understand how den works, nor do I have the time to learn it to the extent that would allow me to "promptly fix problems/bugs caused by" my "contribution." I hope you can understand this: I just *play around* with nix. I do not want someone else who RELIES on it having issues because of my tinkering

also typo in "responsable" lol

i also have NO idea what this means

> am myself legally accountable for it

to what extent would I be "legally" accountable for this contribution? anything that touches the realms of "legality" would be an instant "nope" for me

I tried fixing the bug with AI, but that's it. I REALLY cannot promise anything else; I hope that you can understand this

if my attempt was not good enough, or does not meet contribution standards (which, per the lasts lines of what you linked to me, it does NOT), then, uh... I guess feel free to close the PR? and then open a new one with this same bug again ofc

vic:
Oh it has nothing to do with you particularly. I've also been using AI, I've tagged my own PRs as ai-assisted. 

The issue with AI generated code is that no AI is responsible for it, but humans can contribute code generated by it to projects which have licensing like Den. I know it is practically impossible for us to know what code was our AI trained with, all I ask is that to the best of our knowledge it is not violating GPL or other licensed code. I know you cannot know it , but we need to be explicit what Den maintainers cannot be responsible for some text they did not generate.

I'm not against AI, not against you. I do use AI, so does @sini, it is practically impossible not to use something using AI today, even our code being exposed publicly is already most surely being used to train AI.

 

> I do not want someone else who RELIES on it having issues because of my tinkering

Are Den maintainers responsible for text generated on their behalf? Because when things fail, we maintainers are asked to solve it. We can choose not to, because we are not under a contract and the LICENSE mentions there's no guarantee, so the AI terms kinda contradict the LICENSE precisely because AI agents skip LICENSE awareness. We have no guarantee that AI respect LICENSEs so who is accountable in the future if some big-evil-corp says you stoled their code and contributed as part of an ai-generated contribution.

Please don't take it bad, if you are not willing to use Den, that's ok, Den already uses AI contributions, we just need to be on clear terms about it. I'm not a lawyer (I'm not even an native english speaker, so there you have typos and bad grammar). I've been using AI to help me fix gramamr at documentation. 


I really appreciate you willing to fix Den, I'm not trying to put barriers all the contrary, you acknowledged from the start your contribution was AI generated and I really appreciate that -- the policy is going along those lines, just make that statement explicit, not trying to hide it.

vic:
The AI policy has nothing to do with this particular PR code nor with you as an individual.

vic:
Also contributions to the AI policy are welcome, I just redacted the first version. Even constitutions get revised.

vic:
Maybe tomorrow we open a discussion about this, I'm not a dictator, I'm trying to make it clear that Den does uses AI contributions as almost any other software now. Given that, shall we even acknowledge it, or simply remove those terms and keep on with being happy doing Nix for Den.

vic:
The statement about legally (is more about the LICENSE) when we used to contribute code without AI we used our brains to try come up with code so we said we "authored" it, some companies call "intellectual-property" something you were paid to create for them. Many code is regarded as intellectual-property and because it is property it surely is subject to law, that is the reason behind me writing "law" there. 

Again, I'm not a lawyer, I don't even like lawyers. All contributions are subject to our LICENSE (human made contrbutions) my questions is more about what happens to non-human contributions.

sini:
Shall I take over this bug then?

kalyanoliveira:
@vic: thank you so much for clarifying, I'm just now starting to understand the implications of AI generated contributions to the sustainability of den as a project

I have to admit that I am _still_ a bit confused on this entire discussion, so, @sini by all means, feel free. 

from what I currently gather, if I don't say "yes, I fully understand the patch that the AI made and I am taking responsibility for it," then there would potential issues with licensing. 

it's obvious to me that I will not be able to say "yes, I understand the patch" because, well... that would imply that I would HAVE to FULLY understand the patch. which I don't, and I probably won't in the foreseeable future: learning stuff takes time. and why the hell would one wait for me to learn things just so that this bug can be fixed, when clearly there's interest in getting it fixed:

> Honestly, I'm happy finding this is a bug because for at least a whole hour, I thought I was deeply misunderstanding something.

> omg thanks for that

it has been fun, but I think its time someone actually competent takes over lol

sini:
Fix is now on main. Let me know if there are any remaining issues. @kalyanoliveira Your policy concerns are valid. We had a lively conversation on matrix, but perhaps opening a issue to continue this discussion would be the best place for it.

PR #426, opened 2026-04-10
fix: applyDeep skips static subs that don't consume parametric ctx

sini:
PR #419 introduced applyDeep to propagate parametric context into bare-result sub-includes. But it called takeFn unconditionally on every sub, including full static aspects whose default functor (withOwn parametric.atLeast) discards owned class configs in a non-static branch. Result: `den.aspects.role._.sub.nixos.x = true` was silently dropped when role was a parametric parent that included its sub-aspect.

Gate the inner re-application on canTake.upTo: a static aspect's default functor has empty functionArgs, so upTo is false and the sub is left alone for the static resolve pass. User-provided provider fns (e.g. `{ host, ... }: { nixos = ...; }`) have host in functionArgs, so upTo fires and their config is materialized. This preserves the #419 fix while restoring pre-#419 behavior for static sub-aspects.

Fixes #423.

### DIFF
--- a/nix/lib/parametric.nix
+++ b/nix/lib/parametric.nix
@@ -1,6 +1,6 @@
 { lib, den, ... }:
 let
-  inherit (den.lib) take;
+  inherit (den.lib) take canTake;
   inherit (den.lib.statics) owned statics isCtxStatic;
 
   # Preserve aspect identity through functor evaluation.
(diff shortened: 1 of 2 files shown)


sini:
```markdown
I, Jason Bowman, explicitly acknowledge the following:
```

Filling this out to show that it's not that scary, I had claude code adapt the test from the bug report for me and write the commit message since it's more descriptive than I can manage. :-)

nix/lib/parametric.nix
```diff
@@ -40,13 +40,13 @@ let
...
       r
       // {
-        includes = map (
-          sub:
-          let
-            sr = takeFn sub ctx;
-          in
-          if sr != { } then sr else sub
-        ) r.includes;
+        # Only recurse into subs whose functor actually consumes this ctx.
```
theutz, line 43: Thanks for this comment. Very helpful!

theutz (approved):

PR #428, opened 2026-04-10
feat(core): add den.lib.strict for disabling freeform types

DylanRJohnston:
## Summary

Adds new a new lib module `den.lib.strict` which when imported into a submodule type (usually via den.schema) disables the freeform type which is prone to subtle bugs with scoping and mistyping problems.

It works out of the box with `den.schema.host` and `den.schema.user`, but the types of aspects and flake didn't import a schema type so I added those too.

I've also added support for asserting errors with `denTest`

### DIFF
--- a/modules/options.nix
+++ b/modules/options.nix
@@ -71,5 +71,6 @@ in
     host.imports = [ den.schema.conf ];
     user.imports = [ den.schema.conf ];
     home.imports = [ den.schema.conf ];
+    aspect.imports = [ den.schema.conf ];
   };
 }
--- a/modules/options.nix
+++ b/modules/options.nix
@@ -71,6 +71,5 @@ in
     host.imports = [ den.schema.conf ];
     user.imports = [ den.schema.conf ];
     home.imports = [ den.schema.conf ];
-    aspect.imports = [ den.schema.conf ];
   };
 }
--- a/nix/denTest.nix
+++ b/nix/denTest.nix
@@ -9,9 +9,20 @@
 }:
 let
   # isolated test, prevent polution between tests.
-  denTest = module: {
-    inherit ((evalDen module).config) expr expected;
-  };
+  denTest =
+    module:
+    let
+      config = (evalDen module).config;
+    in
+    {
+      expr = config.expr;
+    }
+    // lib.optionalAttrs (config.expected != null) {
+      expected = config.expected;
+    }
+    // lib.optionalAttrs (config.expectedError != null) {
+      expectedError = config.expectedError;
+    };
 
   # emulate fake-parts only for self and nixpkgs.
   withSystem =
@@ -43,7 +54,28 @@ let
   testModule = {
     imports = [ inputs.den.flakeModule ];
     options.expr = lib.mkOption { };
-    options.expected = lib.mkOption { };
+    options.expected = lib.mkOption { default = null; };
+    options.expectedError = lib.mkOption {
+      type = lib.types.nullOr (
+        lib.types.submodule {
+          options.type = lib.mkOption {
+            type = lib.types.enum [
+              "RestrictedPathError"
+              "MissingArgumentError"
+              "UndefinedVarError"
+              "TypeError"
+              "Abort"
+              "ThrownError"
+              "AssertionError"
+              "ParseError"
+              "EvalError"
+            ];
+          };
+          options.msg = lib.mkOption { type = lib.types.str; };
+        }
+      );
+      default = null;
+    };
     config = {
       den.schema.user.classes = lib.mkDefault [ "homeManager" ];
       den.default.nixos.system.stateVersion = lib.mkDefault "25.11";
--- a/nix/denTest.nix
+++ b/nix/denTest.nix
@@ -17,10 +17,10 @@ let
     {
       expr = config.expr;
     }
-    // lib.optionalAttrs (config.expected != null) {
+    // lib.optionalAttrs (!(config.expected ? undefined)) {
       expected = config.expected;
     }
-    // lib.optionalAttrs (config.expectedError != null) {
+    // lib.optionalAttrs (!(config.expectedError ? undefined)) {
       expectedError = config.expectedError;
     };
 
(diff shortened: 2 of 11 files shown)


nix/lib/aspects/types.nix
```diff
@@ -80,7 +80,10 @@ let
       { name, config, ... }:
       {
         freeformType = lib.types.lazyAttrsOf lib.types.deferredModule;
-        imports = [ (lib.mkAliasOptionModule [ "_" ] [ "provides" ]) ];
+        imports = [
+          (lib.mkAliasOptionModule [ "_" ] [ "provides" ])
+          (if den.schema ? aspect then den.schema.aspect else { })
```
DylanRJohnston: I couldn't find where, if anywhere, the other schema types are initialised, so I made it conditional. Would like to improve this.

nix/lib/strict.nix
```diff
@@ -0,0 +1,4 @@
+{ lib, ... }:
+{
+  _module.freeformType = lib.mkForce null;
+}
```
DylanRJohnston, line 31: This is where the magic happens, surprisingly simple.

nix/lib/types.nix
```diff
@@ -29,6 +29,11 @@ let
         imports = [ den.schema.host ];
         config._module.args.host = config;
         options = {
+          strict = lib.mkOption {
+            description = "";
+            default = false;
+            type = lib.types.bool;
+          };
```
DylanRJohnston: TODO: Remove this. Accidentally left this in from my first attempt at letting the host schema value control it, unfortunately this yields infinite recursion errors.

nix/denTest.nix
```diff
@@ -9,9 +9,20 @@
...
+    in
+    {
+      expr = config.expr;
+    }
+    // lib.optionalAttrs (config ? expected && config.expected != null) {
+      expected = config.expected;
+    }
+    // lib.optionalAttrs (config ? expectedError && config.expectedError != null) {
+      expectedError = config.expectedError;
+    };
```
DylanRJohnston: `nix-unit` has support for asserting errors but we need to expose it. Unfortunately because this passes through a module option first it's a bit tricky as null and not defined are two different things (kinda wish nix had something like `undefined` from javascript as a value). So we need to check if they're defined and not null, and then add it.

nix/denTest.nix
```diff
@@ -43,8 +54,30 @@ let
...
+              "MissingArgumentError"
+              "UndefinedVarError"
+              "TypeError"
+              "Abort"
+              "ThrownError"
+              "AssertionError"
+              "ParseError"
+              "EvalError"
+            ];
+          };
```
DylanRJohnston: Do we care about typing it this strictly?

nix/denTest.nix
```diff
@@ -43,8 +54,30 @@ let
...
+          options.msg = lib.mkOption { type = lib.types.str; };
+        }
+      );
+      default = null;
+    };
     config = {
+      _module.check = false;
```
DylanRJohnston: TODO: Remove this.

nix/flakeOutputs.nix
```diff
@@ -50,6 +50,7 @@ let
     "packages"
     "apps"
     "checks"
+    "tests"
```
DylanRJohnston, line 53: Added a module for exporting `tests`

nix/flakeOutputs.nix
```diff
@@ -60,8 +61,10 @@ let
     };
 
   manySubmodule =
-    lib:
+    lib: imports:
     lib.types.submodule {
+      inherit imports;
```
DylanRJohnston, line 66: This is how we sneak in overrides from `den.schema.flake`

nix/flakeOutputs.nix
```diff
@@ -70,23 +73,25 @@ let
...
 
   flakeTop =
-    lib:
+    { lib, den }:
     lib.mkOption {
       default = { };
       defaultText = lib.literalExpression "{ }";
-      type = (manySubmodule lib);
+      type = (
+        manySubmodule lib [ (if den ? schema && den.schema ? flake then den.schema.flake else { }) ]
```
DylanRJohnston: If only nix had null coalescing like JS `den?.schema?.flake` would be so nice

templates/ci/modules/features/strict.nix
```diff
@@ -0,0 +1,102 @@
...
+        ];
+        flake.arbitray = "value";
+
+        expr = config.flake.arbitray;
+        expectedError = {
+          type = "ThrownError";
+          msg = "The option `flake.arbitray' does not exist";
+        };
+      }
+    );
```
DylanRJohnston, line 97: This test arguably belongs in its own test suite as it's testing the new `den.schema.flake` functionality rather than `den.lib.strict`. Although I guess it is testing that the two play nice together.

templates/ci/modules/test-support/eval-den.nix
```diff
@@ -2,6 +2,6 @@
 {
   imports = [
     inputs.den.flakeModules.denTest
-    inputs.den.flakeOutputs.flake
+    inputs.den.flakeOutputs.tests
```
DylanRJohnston, line 10: We should use the new narrower `tests` module rather than the broad `flake` module.

nix/flakeOutputs.nix
theutz: Can't you do something like `den.schema.flake or {}`?

nix/denTest.nix
theutz: I like that, personally.

nix/lib/aspects/types.nix
vic: Use `den.schema.aspect or { }`. 

I like this! having a base deferred module for all aspects.

Other schemas defined are at `/modules/options.nix`  I know, the filename was originally about what options were available at our entities, but thanks to Dendritic we can rename the file or split without issues.

nix/lib/strict.nix
vic: Awesome!

nix/denTest.nix
vic: lgtm, nicely done

nix/denTest.nix
vic: +1 on having this list

vic:
Wow! Absolute wonderful work, @DylanRJohnston. Please continue with this. Changes make sense to me.

sini:
I, too, like this.

modules/options.nix
```diff
@@ -71,5 +71,6 @@ in
     host.imports = [ den.schema.conf ];
     user.imports = [ den.schema.conf ];
     home.imports = [ den.schema.conf ];
+    aspect.imports = [ den.schema.conf ];
```
DylanRJohnston: A little unsure of this, as aspect is a kind of super type here, not an 'entity' like the others unsure if it should inherit from conf. There's a tension here between just having a spot like `den.default` where it applies to everything, or coming up with the "right" class hierarchy which feels like maybe a trap? Putting `flake` definitely feels wrong through because it's not an aspect at all.

nix/lib/strict.nix
```diff
@@ -0,0 +1,31 @@
...
+            STRICT MODE
+
+            Attempted to set the option "${decl.name}" in "${lib.join "." path}" but no explicit definition exists. If this wasn't a mistake, disable STRICT mode or configure an option. e.g.
+
+            den.schema.${kind}.options.${decl.name} = lib.mkOption { ... };
+
+            See https://documentation.example
+          ''
+        );
+    };
```
DylanRJohnston, line 29: I had this idea last night falling asleep that we could probably hook into some part of the NixOS module resolution system via types that would have enough metadata context to provide errors that point the user towards the fix (inspired by Rust). I think it's pretty cool, let me know what you think!

templates/ci/modules/features/adapter-propagation.nix
```diff
@@ -132,8 +132,8 @@
       {
         den.hosts.x86_64-linux.igloo = { };
 
-        expr = (den.ctx.host { host = den.hosts.x86_64-linux.igloo; }).meta.adapter;
-        expected = null;
+        expr.value = (den.ctx.host { host = den.hosts.x86_64-linux.igloo; }).meta.adapter;
+        expected.value = null;
```
DylanRJohnston: Since we're using `null` as a sentinel value to smuggle the type union `{ expr } & ({ expected } | { expectedErr })` through the module options to be able to surface it to `nix-unit` it becomes impossible to write a test that expects `null`. A simple fix is nesting it inside an attrset, but perhaps we should use a different sentinel value as the default for these options to then filter them out when surfacing the test to `nix-unit`? Could be an arbitrary string that would never get used in a real test or something? Shame nixos doesn't have the ability to define new nominal types or has something like `Symbol` from JS.

templates/ci/modules/test-support/eval-den.nix
```diff
@@ -1,10 +1,21 @@
...
   ];
-  flake = {
+
+  options.flake = {
+    denTest = lib.mkOption { };
+    den = lib.mkOption { };
+  };
```
DylanRJohnston, line 16: This is now required because of the additional of `den` and `denTest` to the flake outputs as of a few hours ago

modules/options.nix
vic: Yeah, maybe it is better to have `den.schema.aspect` but not making it include `den.schema.base`, maybe we should actually rename `den.schema.base` to `den.schema.entity` since it is for them. 

IF we did not include `den.schema.base` at `den.schema.aspect` would be loosing something for strict ?

nix/lib/strict.nix
vic: +1 definitely having errors that can point users on what to do, not only what happened is very good. We should revisit error-messages more on Den later.

templates/ci/modules/features/adapter-propagation.nix
vic: how about expr and expected options to have a default `{ undefined = {}; }`, and test  for `x ? undefined` to see if it is the sentinel. ?

templates/ci/modules/test-support/eval-den.nix
vic: Yep, thanks

modules/options.nix
DylanRJohnston: It just means you have to import strict mode in more places, which IMO isn't that bad. Could also easily export a module `den.strict` or `den.flakeModules.strict` which sets them in all the required places.

vic: importing den.flakeModules.strict to enable on all places looks good to me.

DylanRJohnston: Done

vic (approved):
feel free to merge on succeed

PR #430, opened 2026-04-10
docs: add "Aspect Fixed Point" section 

### LINKED ISSUE #341: document aspect fixed point
also in provides rename the fixedpoint `aspects` to `provides`

drupol:
Fix #341

### DIFF
--- a/docs/src/content/docs/guides/configure-aspects.mdx
+++ b/docs/src/content/docs/guides/configure-aspects.mdx
@@ -285,3 +285,33 @@ And then, you can refer to `tux` aspect from other aspects:
   };
 }
 ```
+
+## Aspect Fixed-Point
+
+Aspects are fixed-point, meaning that they can reference themselves, as such:
+
+```nix
+{
+  den.aspects.igloo =
+    {config,...}: # Here, `config` represents the aspect `igloo` itself.
+    {
+      meta = {
+        default-key = "Hello World!";
+      };
+
+      homeManager = {
+        programs = {
+          gpg = {
+            enable = true;
+            settings.default-key = config.meta.default-key;
+          };
+
+          git = {
+            enable = true;
+            settings.signing.key = config.meta.default-key;
+          }
+        };
+      };
+  };
+}
+```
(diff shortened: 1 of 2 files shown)


docs/src/content/docs/guides/configure-aspects.mdx
```diff
@@ -276,11 +277,43 @@ And then, you can refer to this custom submodule from other aspects:
...
+        default-key = "Hello World!";
+      };
+
+      homeManager =
+        { config, ... }:
+        {
+          programs = {
+            gpg = {
+              enable = true;
+              settings.default-key = config.meta.default-key;
```
theutz: Maybe I'm misunderstanding, but wouldn't the `config` from the `homeManager` class shadow the `config` from the aspects here? Or would `meta` be merged down into the homeManager config like that?

drupol: Nicely spotted, bad copy paste :) thank you, fixed now !

docs/src/content/docs/guides/configure-aspects.mdx
```diff
@@ -276,11 +277,41 @@ And then, you can refer to this custom submodule from other aspects:
...
 ```
+
+## Aspect Fixed-Point
+
+Aspects are fixed-point, meaning that they can reference themselves, as such:
+
+```nix
+{
+  den.aspects.igloo =
+    {config,...}: # Here, `config` represents the aspect `igloo` itself.
```
vic, line 296: If you look at flake-aspects original code, we had a `_modules.args.aspect = config` alias. But this was removed in Den during cleanup since we had no test using that (it is tested in flake-aspects original code). 

Do you think having that alias might be helpful ? To avoid clashes with other `config` from inner dendritic modules like nixos/homeManager ?

@drupol @theutz original code here: https://github.com/vic/flake-aspects/blob/main/nix/types.nix#L88 line 88.

vic: If so, feel free to bring it back at aspects/types. Just add a test using it, so it is no dead code.

drupol: > Do you think having that alias might be helpful ? To avoid clashes with other `config` from inner dendritic modules like nixos/homeManager ?

I don't think so, it is already possible to do it with nix (e.g. `aspect@{ config, ... }:`)

vic:
Thanks @drupol , I'm merging this.

vic (approved):

Issue #429, opened 2026-04-10
BUG: parametrized aspect no longer working


drupol:
Hello,

I'm in the process of migrating my config to `vic/den`.

The pull request is at https://github.com/drupol/infra/pull/124.

From time to time, I update my system (host `x1c`) to get the latest `nixpkgs` update, but also to check if my upcoming branch is working correctly.

Yesterday, I tried to do that because I noticed new commits on `vic/den` project, I got this output in my terminal:

```
┏━ Dependency Graph:
┃ ┌─ ✔ boot.json
┃ │     ┌─ ✔ system_fish-completions
┃ │     │  ┌─ ✔ unit-dbus.service
┃ │     ├─ ✔ user-units
┃ │     │           ┌─ ✔ hm_plasmamanagerdatadesktop_script_wallpaper_picture.js
┃ │     │        ┌─ ✔ home-manager-files
┃ │     │     ┌─ ✔ home-manager-generation
┃ │     │  ┌─ ✔ unit-home-manager-pol.service
┃ │     │  ├─ ✔ unit-accounts-daemon.service
┃ │     │  │  ┌─ ✔ X-Restart-Triggers-polkit
┃ │     │  ├─ ✔ unit-polkit.service
┃ │     │  │        ┌─ ✔ system-path ⏱ 7s
┃ │     │  │     ┌─ ✔ dbus-1
┃ │     │  │  ┌─ ✔ X-Restart-Triggers-dbus
┃ │     │  ├─ ✔ unit-dbus.service
┃ │     ├─ ✔ system-units
┃ │  ┌─ ✔ etc
┃ ├─ ✔ activate
┃ ✔ nixos-system-x1c-26.05.20260409.4c1018d
┣━━━ Builds
┗━ ∑ ⏵ 0 │ ✔ 19 │ ⏸ 0 │ Finished at 11:21:53 after 31s
<<< /run/current-system
>>> /nix/store/v4wgfxyf6cih86ip4l8slnh2iydwwv2j-nixos-system-x1c-26.05.20260409.4c1018d

REMOVED
[R.] alsa-firmware                  1.2.4-zstd
[R.] hwinfo                         25.2, 25.2-lib
[R.] ipw2200-firmware               3.1-zstd
[R.] libreelec-dvb-firmware         1.5.0-zstd
[R.] libx86emu                      3.7
[R.] linux-firmware                 20260309-zstd
[R-] nixos-facter                   0.4.3
[R.] rt5677-firmware                4.16-10-zstd
[...]

PR #432, opened 2026-04-10
BUG: parametrized aspect no longer working

vic:
Adapted from gist at #429


sini:
PR #437 fixes this -- this test doesn't target home-manager (oops) :)

vic:
thanks

PR #449, opened 2026-04-13
feat(core): add typing of core aspect keys in strict mode

DylanRJohnston:
Follow up from #428, when I actually went to use this in my config I found that all the core aspect classes were missing explicit typing.

### DIFF
--- a/modules/aspects/provides/os-class.nix
+++ b/modules/aspects/provides/os-class.nix
@@ -30,4 +30,8 @@ let
 in
 {
   den.ctx.default.includes = [ os-class ];
+  den.schema.aspect.options.os = lib.mkOption {
+    type = lib.types.deferredModule;
+    default = { };
+  };
 }
--- a/nix/flakeOutputs.nix
+++ b/nix/flakeOutputs.nix
@@ -18,7 +18,7 @@ let
       }
       ```
 
-    Same applies to `homeConfigurations`, `packages`, or any other 
+    Same applies to `homeConfigurations`, `packages`, or any other
     flake output attribute that might need merge-semantics for
     multiple values.
 
@@ -35,7 +35,7 @@ let
       }
       ```
 
-    See also flake-parts related error messaage: 
+    See also flake-parts related error messaage:
     https://github.com/hercules-ci/flake-parts/blob/main/modules/flake.nix
   '';
 
@@ -114,7 +114,7 @@ let
     }) systemNames
   );
 
-  all.includes = builtins.attrValues (flakeBased // systemBased);
+  all.imports = builtins.attrValues (flakeBased // systemBased);
 
   flake =
     { lib, config, ... }:
--- a/nix/lib/namespace-types.nix
+++ b/nix/lib/namespace-types.nix
@@ -10,6 +10,8 @@ let
       inherit (den.lib.ctxTypes nsCtxApply) ctxTreeType;
     in
     {
+      imports = [ (den.schema.namespace or { }) ];
+
       options.ctx = lib.mkOption {
         description = "namespace context pipeline";
         defaultText = lib.literalExpression "{ }";
--- a/nix/lib/strict.nix
+++ b/nix/lib/strict.nix
@@ -14,24 +14,66 @@
               lib.head
             ];
 
-            optionPath =
+            explanation =
               if lib.lists.hasPrefix [ "flake" ] path then
-                "den.schema.flake.options.${decl.name}"
-              else if lib.lists.hasPrefix [ "den" "default" ] path then
-                "den.schema.aspect.options.${decl.name}"
+                ''
+                  Attempted to set the flake output option "${decl.name}" but no definition exists.
+                  If this wasn't a mistake, disable STRICT mode, import the flake output or configure an option. e.g.
+
+                  # Import the flake output from den
+                  imports = [ inputs.den.flakeOutputs.${decl.name} ];
+
+                  # Import all flake outputs from den
+                  imports = [ inputs.den.flakeOuputs.all ];
+
+                  # Configure a custom flake output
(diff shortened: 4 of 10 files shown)


nix/lib/strict.nix
```diff
@@ -14,7 +14,13 @@
...
             ];
 
-            kind = if (lib.head path) == "flake" then "flake" else lib.elemAt path 1;
+            kind =
+              if (lib.elemAt path 0) == "flake" then
+                "flake"
+              else if (lib.elemAt path 1) == "default" then
+                "aspect"
+              else
+                lib.elemAt path 1;
```
DylanRJohnston: This "kind" detection is a bit sloppy, I wonder if there's a better way to track schema providence

modules/aspects/provides/os-class.nix
```diff
@@ -30,4 +30,8 @@ let
 in
 {
   den.ctx.default.includes = [ os-class ];
+  den.schema.aspect.options.os = lib.mkOption {
+    type = lib.types.deferredModule;
+    default = { };
+  };
```
DylanRJohnston, line 36: Not sure if it's possible, but it would be cool to only include the schema if the provider is being used. I think the only way to do it would be to turn it into a flakeModule?

nix/flakeOutputs.nix
```diff
@@ -114,7 +114,7 @@ let
     }) systemNames
   );
 
-  all.includes = builtins.attrValues (flakeBased // systemBased);
+  all.imports = builtins.attrValues (flakeBased // systemBased);
```
DylanRJohnston, line 117: This was broken, it's a module not an aspect, so it needs to be imports not includes

nix/strict.nix
```diff
@@ -1,8 +1,23 @@
...
+let
+  moduleOption = lib.mkOption {
+    type = lib.types.deferredModule;
+    default = { };
+  };
+
+  defaultClasses.options = {
+    nixos = moduleOption;
+    darwin = moduleOption;
+  };
```
DylanRJohnston, line 11: Didn't know where to stick these base aspect class definitions, I don't think they should stay here.

templates/ci/modules/features/strict.nix
```diff
@@ -96,7 +96,128 @@
...
+          den.aspects.igloo.wsl.defaultUser = "igloo";
+
+          expr =
+            if inputs ? nixos-wsl then
+              igloo.wsl.defaultUser
+            else
+              lib.warn "nixos-wsl not found in inputs, skipping test `strict-mode.flakeModule.test-aspect-wsl" "igloo";
+          expected = "igloo";
+        }
+      );
```
DylanRJohnston, line 219: `nixos-wsl` isn't in our imports and `nix-unit` doesn't have a concept of "skipping a test for a reason", so this is the nearest thing I could come up with.

nix/flakeOutputs.nix
vic: My bad, thanks.

templates/ci/modules/features/strict.nix
vic: Since `nixos-wsl` is not at inputs, we set `host.wsl.module` to a deferredModule that  defines only `wsl.enable` and `wsl.defaultUser`, see files `features/*wsl*` for tests.

modules/aspects/provides/os-class.nix
vic: I don't think it is possible right now, aspects/providers cannot influence higher levels schemas, I'd like to know if someone finds a way to have that. could enable things like an aspect setting a `host` entity attribute if possible. 


Do we need to make this `options.os` presence optional here? If so I guess you are right about flakeModules.

nix/lib/strict.nix
```diff
@@ -14,16 +14,66 @@
               lib.head
             ];
 
-            kind = if (lib.head path) == "flake" then "flake" else lib.elemAt path 1;
-          in
-          throw ''
-            STRICT MODE
+            explanation =
+              if lib.lists.hasPrefix [ "flake" ] path then
```
vic, line 18: I think using these ifs is the right way to do this, but I dont quite like large if branches, I'd prefer to have messages assigned as let bindings and have a short if chain at the end referencing those messages, that we can have visibility of what the if condition is doing.

nix/strict.nix
vic: We can move those to /modules, those classes are very common, and we will likely always have those definitions loaded, so just create a file for them at modules.

nix/lib/namespace-types.nix
```diff
@@ -10,6 +10,8 @@ let
       inherit (den.lib.ctxTypes nsCtxApply) ctxTreeType;
     in
     {
+      imports = [ (den.schema.namespace or { }) ];
```
vic, line 13: 👍

vic:
Hey @DylanRJohnston we have finally merged The Big fx-pipeline branch #475. Do you want me to rebase and solve conflicts so we can keep progress on this ?

sini:
Hello @DylanRJohnston I could use your review on:
https://github.com/sini/den-schema/ => https://github.com/sini/den-schema/tree/main/templates/demo

Will this satisfy your needs if we swap the current entity system for it?

Also any review feedback you have on the design or implementation is 100% welcome.

DylanRJohnston:
Apologies @sini and @vic work has gotten really busy these last few weeks and I haven't had the chance to come back and finish this work off. I'd say you can close this PR for now although strict mode isn't very useful without it. I'll have to fix it when I get around to updating my config

sini:
@DylanRJohnston no worries. If you have time to look at den-schema, it's a ground-up clean-room re-implementation of den's schema library that we're going to be subbing in and I believe it should cover your use-case requirements. 

Soliciting review to confirm that. :)

PR #452, opened 2026-04-14
fix: make home hostName nullable for standalone homes

sini:
Standalone home-manager configurations may not have a bound host. Change hostName from strOpt to nullOr str so it defaults to null when no host is associated.


vic (approved):
FFTM (Feel free to merge) +1

PR #453, opened 2026-04-14
feat: context stage tagging + __ctxTrace on ctxApply output

sini:
tagStage marks includes with __ctxStage, __ctxKind, __ctxAspect for tracing which context stage produced each include.

traceItem produces serializable summaries of traversal items with key, selfName, prevName, provider flags, ctxKeys, entityNames.

ctxApply now emits __ctxTrace alongside includes — a list of traceItem records for downstream consumers (diag pipeline, structuredTrace).

### DIFF
--- a/nix/lib/ctx-apply.nix
+++ b/nix/lib/ctx-apply.nix
@@ -77,17 +77,39 @@ let
     in
     [ args ] ++ lib.concatMap expandOne intoList;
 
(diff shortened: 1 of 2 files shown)


vic:
deferred until we merge fx

sini:
I'll sneak this in with the massive FX + diagram merge

PR #454, opened 2026-04-14
feat: create battery for flake-parts custom perSystem classes

theutz:
## TODO

- [ ] Write tests
- [ ] Write docs
- [ ] Add batteries for the custom classes as sub-aspects?


theutz:
@vic Do you think it would be worth creating separate batteries for the examples that you had in the template as well? Or should we encourage people to define their own custom classes?

theutz:
Not sure this is relevant as-is given the rewrite. Closing for now!

PR #458, opened 2026-04-15
test: bogus to showcase bug in aspect custom submodules

theutz:
I might just be misunderstanding how to use this, but I followed the instructions [here](https://den.oeiuwq.com/v0.16.0/guides/configure-aspects/#aspect-custom-submodule). I've been having trouble using `imports` on any aspect, actually, but in different failure modes. This is just one I've been able to isolate.

### DIFF
--- a/templates/bogus/modules/bug.nix
+++ b/templates/bogus/modules/bug.nix
@@ -6,18 +6,18 @@
       {
         den,
         lib,
-        igloo, # igloo = nixosConfigurations.igloo.config
+        # igloo, # igloo = nixosConfigurations.igloo.config
+        apple,
         tuxHm, # tuxHm = igloo.home-manager.users.tux
         ...
       }:
       {
-        # replace <system> if you are reporting a bug in MacOS
-        den.hosts.x86_64-linux.igloo.users.tux = { };
+        den.hosts.aarch64-darwin.apple.users.tux = { };
 
         # do something for testing
         den.aspects.tux.user.description = "The Penguin";
 
-        expr = igloo.users.users.tux.description;
+        expr = apple.users.users.tux.description;
         expected = "The Penguin";
       }
     );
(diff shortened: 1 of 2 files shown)


kalyanoliveira:
I don't think that [72ebb63](https://github.com/vic/den/pull/458/commits/72ebb63527dc0ffb95fd2e0bbee0da303fdde36b) quite reflects what https://den.oeiuwq.com/guides/configure-aspects/#aspect-custom-submodule shows?

Cause in https://den.oeiuwq.com/guides/configure-aspects/#aspect-custom-submodule, what we see is an `imports` attribute under an aspect:

```nix
den.aspects.foo.imports = { ... };
```

And what you have in [72ebb63](https://github.com/vic/den/pull/458/commits/72ebb63527dc0ffb95fd2e0bbee0da303fdde36b) is an `imports` attribute under a user's schema?

```nix
den.hosts.aarch64-darwin.apple.users.tux.imports = { ... };
```

Would you be able to elaborate on what exactly was your initial intention with [72ebb63](https://github.com/vic/den/pull/458/commits/72ebb63527dc0ffb95fd2e0bbee0da303fdde36b)?

templates/bogus/modules/bug.nix
```diff
@@ -12,13 +12,28 @@
...
+                    };
+                  };
+                }
+              ];
+            };
+        };
 
-        expr = apple.users.users.tux.description;
-        expected = "The Penguin";
+        expr = apple.users.users.tux.categories.programs;
```
vic, line 35: you are testing apples and oranges here, :D

apple is the instantiated darwinConfiguration, its users.users.tux does not have categories.

you explicitly added `categories` here: `den.hosts.aarch64-darwin.apple.users.tux.categories` that is, the user entity.

vic: values from user-entity are not applied directly into nixos/darwin configurations. entities are for description of infra entities. based on values from those, aspects provide configurations.

PR #463, opened 2026-04-17
fix: preserve unresolvable includes in applyDeep

### LINKED ISSUE #460: Parametized Aspects incorrectly de-duplicated when parameter isn't a context entity
**Description**
There is a regression in `v0.16.0` where parametrized aspects that are not fed by the context pipeline (such as the `unfree` battery) are aggressively and incorrectly de-duplicated, preventing them from evaluating properly when called multiple times with different arguments.

Originally reported via discussion [#459](https://github.com/vic/den/discussions/459).

**Root Cause**
We recently updated aspects to give them a proper identity based on their declaration path and name. When the tree of all aspects is expanded via their `includes`, we de-duplicate them based on this identity (e.g., if `desktop` includes `zsh` and `develop` includes `zsh`, `zsh` is only pulled in once).

The problem arises with parametrized aspects. If we consider an aspect as a function `F`, de-duplication works perfectly. However, a call like `den._.unfree [ "drawio" ]` is actually `F(X)`. If another aspect includes `den._.unfree [ "steam" ]`, that is `F(Y)`. 

Currently, the de-duplication logic identifies both `F(X)` and `F(Y)` simply as `F`. It strips the parameters from the identity check, meaning only the very first `unfree` invocation survives de-duplication, and all subsequent usages are silently dropped.

**Steps to Reproduce**
1. Create a configuration that calls `den._.unfree` (or another non-context parameterized aspect) from at least two different aspects with different arguments.
2. Attempt to evaluate or build.
3. Observe that only the first package is permitted, and Nix throws an unfree license evaluation error for the second.

*(A minimal reproducible flake is pending from the original reporter, but this can likely be reproduced locally based on the above conditions).*

**Proposed Resolution**
Update the de-duplication logic to correctly distinguish and preserve parametrized aspects. The identity check needs to account for the arguments being passed, so that `F(X)` and `F(Y)` are treated as distinct inclusions. Due to the way aspect identity is currently tracked, this may require a deeper structural fix rather than a simple patch.

sini:
## Summary
- `applyDeep` eagerly replaced sub-aspect includes it couldn't resolve with the current context (`{host,user}`) with `{}`, destroying parametric includes like `den._.unfree` that need `{class, aspect-chain}`
- Preserve the original include when `applyDeep` returns `{}` so the statics path can resolve it with the correct context

## Test plan
- [x] `deadbugs-issue-460` tests: parametric wrapper around unfree calls on both fx and legacy pipelines
- [x] Full CI suite passes (491/491)

Fixes #460

### DIFF
--- a/templates/ci/modules/features/deadbugs/issue-460-parametric-dedup.nix
+++ b/templates/ci/modules/features/deadbugs/issue-460-parametric-dedup.nix
@@ -1,0 +1,109 @@
+# Issue #460: Parametric aspects with non-context args (like unfree) were
+# silently dropped when included via a parametric wrapper function.
+#
+# Root cause: applyDeep eagerly replaced includes it couldn't resolve with
+# the current context ({host,user}) with {}, destroying them before the
+# statics path could resolve them with {class, aspect-chain}.
+{ denTest, ... }:
+{
+  flake.tests.deadbugs-issue-460 = {
+
+    # Baseline: static includes work (never broken).
+    test-unfree-static-includes = denTest (
+      {
+        den,
+        lib,
+        tuxHm,
+        ...
+      }:
+      {
+        den.hosts.x86_64-linux.igloo.users.tux = { };
+
+        den.aspects.apps._.app-a.includes = [ (den._.unfree [ "drawio" ]) ];
+        den.aspects.apps._.app-b.includes = [ (den._.unfree [ "obsidian" ]) ];
+
+        den.aspects.tux.includes = [
+          den.aspects.apps._.app-a
+          den.aspects.apps._.app-b
+        ];
+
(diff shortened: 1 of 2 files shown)


templates/ci/modules/features/deadbugs/issue-460-parametric-dedup.nix
```diff
@@ -0,0 +1,109 @@
...
+
+        expr = lib.sort (a: b: a < b) tuxHm.unfree.packages;
+        expected = [
+          "drawio"
+          "obsidian"
+        ];
+      }
+    );
+
+    # Same regression on legacy pipeline.
```
vic, line 73: awesome, having both tests for now is good(tm)

vic (approved):

PR #466, opened 2026-04-17
feat: host-aspects battery for projecting host homeManager to users

sini:
## Summary
- New battery `den._.host-aspects` that projects all homeManager-class configs from the host's aspect tree onto users who opt in
- Wraps `host.aspect` with `fixedTo { host, user }` so the pipeline resolves it with `class = "homeManager"`, collecting only homeManager keys
- Other class keys (nixos, darwin) are ignored naturally by the pipeline

## Usage

```nix
den.aspects.tux.includes = [ den._.host-aspects ];
```

## Test plan
- [x] Host aspect with homeManager key projects to opted-in user
- [x] Host aspect with only nixos key does not leak into user's homeManager
- [x] Multiple host sub-aspects with homeManager keys all project
- [x] User who does NOT include the battery does not receive host homeManager configs
- [x] No circular evaluation when accessing both host and user configs
- [x] Full CI suite passes (498/498)


vic:
Thanks! I will merge.

My plan for the future (and for solving Pol's blogpost) is that we can make our fx handlers be much better at deduplication, discarding double contributions (this is related to you hashing arg values).

Once that is real, we won't need mutual-provider nor this battery. But we will also need auto-activated capabilities.

vic (approved):

PR #468, opened 2026-04-17
fix: module-level dedup for host-aspects overlap support

sini:
Named aspects now get a key attribute (class@identity) on their collected modules. The NixOS module system deduplicates by key across independent resolve calls, so a user can include an aspect directly AND receive it via host-aspects without duplicate option declarations.

Anonymous/synthetic aspects are excluded from keying so multiple anonymous includes coexist correctly.

Also fixes the pre-existing duplication where aspects included by both host and user produced duplicate nixos modules via the default context transition.

### DIFF
--- a/modules/aspects/provides/host-aspects.nix
+++ b/modules/aspects/provides/host-aspects.nix
@@ -4,7 +4,7 @@ let
 
   description = ''
     Projects all homeManager-class configs from the host's aspect tree
-    onto users who opt in.
+    onto users who opt in. Requires the fx pipeline.
 
     ## Usage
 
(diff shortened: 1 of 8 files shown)


drupol (approved):
This PR fix the issue I described in https://not-a-number.io/2026/evaluating-den-a-dendritic-configuration-framework/ !!!

Looking forward for it... 

GG !!!

theutz (commented):
Oooh I hope this works! 🤞

theutz (approved):

vic:
Awesome @sini! Thanks for this! 


With this working now -and fixing @drupol post- about duplication I believe we can bring back the builtin bidirectionality (that we removed because our legacy core was not able to do this dedup enabled in this fix). 

What I mean by bidirectionality:

That host-aspects correctly contribute user classes `homeManager`, `hjem` to all its users without people having to use any battery. 

We used to do that with this line: https://github.com/vic/den/pull/272/changes#diff-292456ccf182ad0171aa662b32637c8c4591f0d7e471b8bdec8753ba6d09e4efL20 

We don't have necessarily to revert that file, but the ideas is that aspects as those defined by drupol's post must work naturally without any extension.

vic:
We can chat about this, I'd like to know what you think about this (because I know you have better plans -- capabilities)

sini:
Yes.

Honestly, I think the pattern you settled on is actually better, we simply need document and educate users about the new host-aspects battery. I have many users I install on my machines who only need slim accounts.

theutz:
I wonder if having a "isNormalUser"/"isSystemUser"-like toggle might help everyone then when they create new users? So the intentions are explicit for each new user?

sini:
> I wonder if having a "isNormalUser"/"isSystemUser"-like toggle might help everyone then when they create new users? So the intentions are explicit for each new user?

Is that not what the aspect include provides? Perhaps it simply needs a better name.

PR #470, opened 2026-04-19
feat: allow host-aspects to work with other user classes

vic:
Iterates over `user.classes` instead of just being fixed to `homeManager` only. Add tests for hjem class.


sini (dismissed):
LGTM

vic:
I'll merge, I'm also using this to test our GH actions now running `just ci` fast on new PR. and leave slower ci-deep for when `allow-ci` is present. So that people can iterate faster on their PRs.

PR #475, opened 2026-04-23
feat: fx effects pipeline — replace legacy resolver

sini:
## Summary

Replace den's legacy recursive resolver with an algebraic effects pipeline (via nix-effects). The entire resolution lifecycle — from policy routing to aspect compilation to module collection — is now driven by composable effect handlers.

- **Remove legacy resolver** — delete adapters.nix, resolve.nix, statics.nix (~450 lines). Fx handlers own everything: include resolution, parametric args, transitions, constraints, dedup.
- **Four-concern separation** — decompose `den.ctx` into `den.policies` (directed entity edges), `den.stages` (scoped behavior bindings), `den.schema` (entity types), and `den.aspects` (reusable behavior).
- **Effectful bootstrap** — pipeline setup is itself effectful via `resolve-policy` and `resolve-target` effects, making the entire lifecycle interceptable.
- **Per-policy named handlers** — policies can declare `handlers` that get `scope.provide`'d into transition sub-computations (Layer 2 foundation).
- **`den.ctx` compat shim** — always-imported, forwards `den.ctx.*` to `den.stages.*` with deprecation warnings. Covers scoped behavior + into forwarding.

**481 tests passing** (22 new). Net: -149 lines across 197 files.

## Detailed walkthrough

[Simplified Review Guide](https://gist.github.com/sini/dea443d26352b3cbd667d88ee4b34508) — plain-language explanation of the architecture, handler vocabulary, context propagation, and what changed from main.

## Commits

| Commit | What |
|--------|------|
| remove legacy resolver | Delete adapters/resolve/statics, gut parametric, fx handlers own resolution |
| ctx-as-data transitions | scope.provide + __scopeHandlers, deferred includes, fan-out identity |
| edge cases | Fan-out dedup, cross-providers, perCtx, constraint cascade |
| parametric type separation | __fn/__args replace __functor, scope.provide replaces scope.stateful |
| four-concern separation | den.stages + den.policies + entity consolidation, remove den.ctx |
| code quality | Decompose oversized functions, naming, comments, efficiency |
| policies rename + compat | den.relationships → den.policies, ctx compat shim |
| effectful bootstrap | resolve-policy/resolve-target effects, policy handlers, depth guards, tests |

## Next steps (separate PRs)

- **aspect-chain removal** — 3 files still use `fromAspect`/`aspect-chain`; replace with `item.resolved`
- **Generic fan-out isolation** — current fix scopes to `targetClass == "flake"`; needs hierarchical dedup
- **Layer 2: per-policy named effects** — the `handlers` field is wired; next is user-facing documentation and batteries that use it
- **Vestigial removal** — `parametric.nix`, `take.nix`, `perHost-perUser.nix` once downstream migrates

## Test plan

- [x] 481/481 CI tests pass (`nix develop -c just ci`)
- [x] Downstream validation against:
  - [theutz/slashfiles](https://github.com/theutz/slashfiles)
  - [drupol/infra](https://github.com/drupol/infra)
  - [sjcobb2022/nix-den](https://github.com/sjcobb2022/nix-den)
  -- all hosts work with ctx compat shim -- even fixed another bug the owner was facing :1st_place_medal: 
- [ ] Review by @vic

## Documentation overhaul
Rendered documentation can be previewed at: https://den.json64.dev/

🤖 Generated with the aid of [Claude Code](https://claude.com/claude-code)


sini:
TODO: 
Docs revamp - changed some things, need another update/rework
Followup work on policies w/ ACL example - need a demo template exercising them in a rich capacity
Diagram generation -- regression, need fixing
~Have some brave users actually test this branch~

TODO more like DONE

sini:
Working on some straggling edge-cases around the flake / flake-packages / flake-system integration. Otherwise the core is ready for review, only minor fixes pending.

sini:
I was hoping to dodge it -- but alas. Larger change coming in replacing forwards with an effect handler. :)

modules/options.nix
```diff
@@ -6,7 +6,15 @@
 }:
 let
   inherit (config) den;
-  types = import ./../nix/lib/types.nix {
+  hostEntities = import ./../nix/lib/entities/host.nix {
```
vic: I saw that `hostEntities` and `homeEntities` both are exposed by `lib/types.nix` access them from `types` instead of re-importing here.

modules/policies/flake.nix
```diff
@@ -0,0 +1,73 @@
...
+    "apps"
+    "checks"
+    "devShells"
+    "legacyPackages"
+  ];
+
+  # A flake-system context has system but not entity-level keys. The existing
+  # ctx into functions use strict { system } destructuring that rejects extra
+  # attrs, so we must avoid firing when host/home are also present.
+  isFlakeSystemCtx = ctx: ctx ? system && !(ctx ? host) && !(ctx ? home);
```
vic: This `isFlakeSystemCtx` and the line below: ` if !(ctx ? host) && !(ctx ? system) && !(ctx ? home) then ` and other similar checks here are somewhat strange to me, if I understand the comment correctly, it seems like they are preventing some kind of bug from firing. 

Why do we have to check the context for not having host / home, etc. WHEN is the context having those keys ? At this flake-level they should never be present, if they are we have a bug, and we should not try to get around by checking ctx keys.

modules/policies/flake.nix
```diff
@@ -0,0 +1,73 @@
...
+        else
+          [ ];
+    };
+
+    flake-system-to-flake-os = {
+      from = "flake-system";
+      to = "flake-os";
+      resolve =
+        ctx:
+        if isFlakeSystemCtx ctx then
```
vic: remove condition. instead replace `ctx: ` by `{ system }: ` since that is the expected context here, being called without that means there's a problem on how policies are dispatched.

modules/policies/flake.nix
```diff
@@ -0,0 +1,73 @@
...
+        else
+          [ ];
+    };
+
+    flake-system-to-flake-hm = {
+      from = "flake-system";
+      to = "flake-hm";
+      resolve =
+        ctx:
+        if isFlakeSystemCtx ctx then
```
vic: same, remove condition, context is the condition: `{ system }: `

modules/policies/core.nix
```diff
@@ -0,0 +1,29 @@
+# Core entity policies — fundamental traversal between entity kinds.
+{ lib, ... }:
+{
+  den.policies = {
+    host-to-users = {
+      from = "host";
+      to = "user";
+      resolve =
+        ctx:
+        if !(ctx ? host) || !(builtins.isAttrs ctx.host) || !(ctx.host ? users) then
```
vic: Remove this conditional, let it crash instead of masking potential problems. The context here is expected to be `{ host }`, change `ctx: ` to `{ host }:`.

vic: using this kind of conditions will silently ignore potential problems.

modules/policies/core.nix
```diff
@@ -0,0 +1,29 @@
...
+        else
+          map (user: {
+            inherit (ctx) host;
+            inherit user;
+          }) (lib.attrValues ctx.host.users);
+    };
+    host-to-default = {
+      from = "host";
+      to = "default";
+      resolve = ctx: if !(ctx ? host) || !(builtins.isAttrs ctx.host) then [ ] else [ ctx ];
```
vic: remove condition, if the policy has `from = "host"` it means the context must have `{ host }`. right ?

modules/policies/core.nix
```diff
@@ -0,0 +1,29 @@
...
+    };
+    user-to-default = {
+      from = "user";
+      to = "default";
+      resolve = ctx: if !(ctx ? user) || !(builtins.isAttrs ctx.user) then [ ] else [ ctx ];
```
vic: use: `resolve = lib.singleton`  the context is un-changed. Why do all these conditionals exist ? Is there any possibility that this policy is called with a context NOT having a user ? If so, that is a bug. Do not workaround with conditions like this.

nix/lib/aspects/fx/handlers/ctx.nix
```diff
@@ -9,6 +9,9 @@
 let
   # Build handler set from context.
   # Each key in ctx becomes a handler that resumes with the value.
+  # has-handler (from nix-effects) replaces probe-arg — it queries
```
vic: what is `probe-arg` ? Remove comments talking about intermediate stuff that was removed.

modules/policies/batteries.nix
```diff
@@ -0,0 +1,116 @@
...
+      };
+    };
+
+    hjem-host-to-hjem-user = {
+      from = "hjem-host";
+      to = "hjem-user";
+      resolve = mkIntoClassUsers "hjem";
+    };
+
+    # --- maid ---
```
vic: Do not have all batteries policies here. Each policy belongs in the same file/directory as its battery. removing that battery file/directory also removed the policy. move all these policies into their correct places.

modules/outputs/osConfigurations.nix
```diff
@@ -1,18 +1,13 @@
...
-
   osFwd =
     { host }:
     den.provides.forward {
       each = lib.optional (host.intoAttr != [ ]) true;
       fromClass = _: host.class;
       intoClass = _: "flake";
       intoPath = _: [ "flake" ];
-      fromAspect = _: den.ctx.host { inherit host; };
+      fromAspect = _: den.lib.resolveStage "host" { inherit host; };
```
vic: Question: Why use resolveStage instead of host.resolved ? It appears that resolved uses resolveStage underneath, have only one way to do things.

modules/context/perHost-perUser.nix
```diff
@@ -1,15 +1,51 @@
...
+  perCtx =
+    requiredKeys: aspect:
+    let
+      reqKeysSorted = builtins.sort builtins.lessThan requiredKeys;
+      extraKeys = builtins.filter (k: !(builtins.elem k reqKeysSorted)) allContextKeys;
+      # Required keys as required (false), extra keys as optional (true)
+      funcArgs = lib.genAttrs reqKeysSorted (_: false) // lib.genAttrs extraKeys (_: true);
+    in
+    lib.warn
+      "den.lib.perCtx [${lib.concatStringsSep "," reqKeysSorted}] is deprecated — handler-based resolution makes context guards unnecessary"
```
vic: Improve error, it is better that error messages communicate people what they have to do: "Remove perCtx" or "Use this instead: ...", they dont care much about the inner workings, but more on how to make their system work.

modules/policies/batteries.nix
```diff
@@ -0,0 +1,116 @@
...
+      };
+    };
+
+    hm-host-to-hm-user = {
+      from = "hm-host";
+      to = "hm-user";
+      resolve = mkIntoClassUsers "homeManager";
+    };
+
+    # --- hjem ---
```
vic: each battery policy is part of the battery, not of this file.

modules/policies/batteries.nix
```diff
@@ -0,0 +1,116 @@
...
+    };
+
+    # --- WSL ---
+    # WSL activation is host-config-driven (wsl.enable), not user-class-driven,
+    # so it doesn't use mkDetectHost which requires matching user classes.
+    host-to-wsl-host = {
+      from = "host";
+      to = "wsl-host";
+      resolve =
+        ctx:
```
vic: I will stop reviewing these `ctx: ` functions, use the expected context `{ host }: ` as arguments instead of using conditions `ctx ? host ...`.

nix/lib/aspects/fx/trace.nix
```diff
@@ -73,41 +73,49 @@ let
...
+          if chainTip == null then
+            null
+          else
+            let
+              segments = lib.splitString "/" chainTip;
+              # Drop {ctxId} suffixes (segments starting with "{").
+              base = builtins.filter (s: builtins.match "\\{.*" s == null) segments;
+            in
+            if base == [ ] then chainTip else lib.last base;
         meaningful =
```
vic: We have other definitions of isMeaningfulName. Maybe use a single one.

nix/lib/aspects/fx/handlers/include.nix
```diff
@@ -97,66 +101,136 @@ let
...
+        probeArgs =
+          keys:
+          if keys == [ ] then
+            fx.pure true
+          else
+            let
+              key = builtins.head keys;
+              rest = builtins.tail keys;
+            in
+            fx.bind (fx.send "has-handler" key) (
```
vic: name `"has-handler"` is nix-effects internal name dont use it, use `hasHandler` API instead.

nix/lib/aspects/fx/aspect.nix
```diff
@@ -6,18 +6,26 @@
...
     "name"
+    "description"
     "meta"
     "includes"
     "provides"
     "into"
+    "__fn"
+    "__args"
     "__functor"
-    "__functionArgs"
```
vic, line 17: Are you sure we want to keep __functor and remove __functionArgs ? Just asking because they their meaning is related one to another.

vic (dismissed):
I did a quick review up to my energy posibilities. Added a few comments, maybe we can also run a code-simplification run on code after forwards are done.

Ping me when I can review again.

modules/policies/core.nix
sini: These patterns are the result of a more complicated issue than you know, and exists because of tests wired up like this:
`ctx = { host = "igloo"; system = "x86_64-linux"; }`

That said, can do, will add coercion or guards around it to try and make this consistent.

There are some other places where there's some extra logic because now that effects fall through to the root, they implicitly widen, wherein in the old pipeline things didn't back-propagate in the same way.

I'll get these coerced into your format.

modules/options.nix
sini: Fixed — using types.nix re-export now.

modules/policies/flake.nix
sini: Added ctxSatisfiesEntity guard in synthesize-policies.nix — dispatch validates entity keys before calling resolve. Policies use clean { host, ... }: now. The flake-system scope exclusion (isFlakeSystemCtx) needs dispatch-level scope exclusion to remove — deferred.

modules/policies/core.nix
sini: Fixed. core.nix uses { host, ... }: and lib.singleton. The dispatch guard in synthesize-policies.nix makes this safe.

modules/policies/batteries.nix
sini: Done — each policy moved to its battery file. batteries.nix deleted.

nix/lib/aspects/fx/handlers/ctx.nix
sini: Removed.

nix/lib/aspects/fx/handlers/include.nix
sini: Fixed — using fx.effects.hasHandler.

nix/lib/aspects/fx/aspect.nix
sini: Added back to structuralKeysSet.

modules/context/perHost-perUser.nix
sini: Improved — tells users what to use instead.

modules/outputs/osConfigurations.nix
sini: Investigated — host.resolved and resolveStage produce different identity, breaks the forward each filter (h != host). Keeping resolveStage for now.

nix/lib/aspects/fx/trace.nix
sini: Checked — single definition in types.nix:24, re-exported and imported in multiple files. Already deduplicated.

nix/lib/aspects/fx/aspect.nix
sini: I want to actually remove it, but there were some awkward edge-cases that necessitated it's reintroduction until they can be more thoroughly investigated.

modules/policies/flake.nix
sini: Investigated the scope guards thoroughly. The issue is that currentCtx in the pipeline accumulates monotonically across all transitions. When flake-system transitions to flake-os (adding host to ctx), subsequent flake-system policies (flake-packages, etc.) see that host in the accumulated ctx. Restoring ctx after each transition breaks everything — sibling transitions legitimately depend on accumulated context (e.g., host-to-users needs host from a parent transition). The isFlakeSystemCtx guard is a valid scoping mechanism until we add proper scope snapshots at transition boundaries. Entity-level guards (host/user/home isAttrs checks) are now handled by the dispatch layer in synthesize-policies.nix.

sini: Fixed in 035540ee. The dispatch layer (synthesize-policies.nix) now handles flake scope: flake-* policies only fire when no entity attrset values are in context. Value type detection, no hardcoded keys. isFlakeSystemCtx removed. All policies use clean { system, ... }: destructuring.

modules/aspects/provides/os-class.nix
```diff
@@ -24,10 +24,18 @@ let
       fromClass = _: "os";
       intoClass = lib.id;
       intoPath = _: [ ];
-      fromAspect = _: lib.head aspect-chain;
+      fromAspect = _: aspect;
+      fromCtx = _: ctx;
```
drupol: A bit unsure about this one, please let me know.

```suggestion
      fromCtx = _: ctx;
      inherit description;
```

modules/aspects/provides/os-user.nix
```diff
@@ -40,11 +40,12 @@ let
         "users"
         user.userName
       ];
-      fromAspect = _: den.lib.parametric.fixedTo { inherit host user; } user.aspect;
+      fromAspect = _: user.aspect;
```
drupol:
```suggestion
      fromAspect = _: user.aspect;
      inherit description;
```

sini: I don't know how valuable adding description here is as it's not used anywhere and just increases the object size.

If this was for the name field and identity, I'd 100% agree. Let's defer until we actually see a usage of description (and then probably put it on meta.description)

modules/outputs/osConfigurations.nix
sini: Leaving this comment open for reference, but I think we opted to defer it for now per our discussion on matrix.

Once we have the multi-pass evaluation for capabilities/provide-to, perhaps worth revisiting.

nix/lib/aspects/fx/aspect.nix
sini: I've got taking another stab at killing __functor as followup work on my tracking list. As it stands, it's usage is greatly reduced and simplified.

modules/aspects/provides/os-user.nix
sini: I'll wait for your response to see if you agree or not before hitting resolve. It costs me nothing to just hit the accept+merge button if you have any strong feelings. :)

drupol: Totally fine mate, go ahead

nix/lib/aspects/types.nix
```diff
@@ -1,117 +1,198 @@
 { lib, den, ... }:
 let
-  inherit (den.lib) lastFunctionTo canTake;
+  inherit (den.lib) canTake;
 
   isSubmoduleFn = canTake.upTo {
```
vic, line 6: This is the only usage of `canTake.*` now, we can replace with `intersectAttrs { lib, config, option } (functionArgs f) != { }` and remove canTake and take.* libs.

vic: Oh, nevermind, I forgot we are in backwards compatibility mode. We cannot remove take.* for now.

nix/lib/synthesize-policies.nix
```diff
@@ -0,0 +1,107 @@
...
+  den,
+  ...
+}:
+let
+  # Infer the required entity key from a stage name.
+  # from = "host" or "hm-host" implies ctx.host must be an attrset entity.
+  # from = "user" or "hm-user" implies ctx.user must be an attrset entity.
+  # from = "home" implies ctx.home must be an attrset entity.
+  # Other stages (flake, flake-system, etc.) have no entity requirement.
+  entityKeyFor =
```
vic: for @vic: I still need to understand how policies work exactly. Why does it appear we have to handle special cases like this. I was expecting policies to be simpler than den.ctx transitions. At least because they are only about transforming data and nothing else.

nix/lib/stage-types.nix
```diff
@@ -0,0 +1,51 @@
...
+          "_module"
+        ];
+        hasKey = x: builtins.any (k: x ? ${k}) stageNodeKeys;
+        isLeaf = lib.any (d: hasKey d.value) coerced;
+      in
+      if isLeaf then
+        stageSubmodule.merge loc coerced
+      else
+        (lib.types.lazyAttrsOf stageTreeType).merge loc coerced;
+    emptyValue = {
```
vic: Remove this `emptyValue`, this is exactly the default mkOptionType has.

nix/lib/resolve-stage.nix
```diff
@@ -0,0 +1,42 @@
...
+      classAttrs = builtins.removeAttrs stageNode structuralKeys;
+      scopeHandlers = constantHandler ctx;
+    in
+    classAttrs
+    // {
+      name = stageNode.name or name;
+      meta =
+        let
+          stageInto = stageNode.meta.into or null;
+          into = den.lib.synthesizePolicies.mergePolicyInto name stageInto;
```
vic: Ok, so this is the place where synthesize-policies.nix is being used. And it looks like it is for `stage.meta.into` which IIRC was for ctx.into backwards compat. 

However, this also sets { meta.into = fn } which looks like an old ctx.into transition. Defeating the purpose of having separation between stages (which are aspects) and policies (transitions -- relationships).

Looks like at the end of the day, stages (aspects) still have meta.into transitions. :/

sini: Addressed in the policy spec PRs (#477 + #478).

In #478, `mergePolicyInto` is removed from `resolve-stage.nix` — stages only carry manual `into` if explicitly declared, otherwise `null`. Policies dispatch via per-policy named effect handlers (`policy:<name>`) in the transition handler, completely separate from the stage aspect.

Clean separation: stages are aspects (behavior), policies are transitions (relationships). They no longer mix at `meta.into`.

nix/lib/synthesize-policies.nix
sini: This is gone on the policies branch as well

nix/lib/resolve-stage.nix
sini: Also: `entityKeyFor` in `synthesize-policies.nix` no longer hardcodes `["host" "user" "home"]`. It derives entity kinds from `den.schema` at eval time, so user-defined entity kinds are automatically supported. Matches the `options.nix` change from `1b5b3b28`.

nix/lib/namespace.nix
```diff
@@ -13,7 +13,15 @@ let
   # both _ and provides with identical content. Re-importing both causes
   # listOf options (like includes) to merge duplicates.
   stripAliases = lib.mapAttrs (
-    _: v: if builtins.isAttrs v then builtins.removeAttrs v [ "_" ] else v
+    _: v:
+    if builtins.isAttrs v then
+      builtins.removeAttrs v [
+        "_"
+        "__functor"
+        "__ctx"
```
vic: Maybe we need more test that exercise namespaces (beyond flake frontier), because here we remove `__ctx`, but if you grep for it, almost all places that use `__ctx` also have `__scopeHandlers` and they are almost identical, I'd prefer if we have only one of those, I personally prefer `__scopeHandlers` only.

There's also `__ctxId`, `__fn`, `__args`, `__parentScopeHandlers`, `__parametricDepth` (and many many others `__*`) and I don't know how they will play with sharing namespaces between flakes.

nix/lib/namespace-types.nix
```diff
@@ -1,20 +1,16 @@
...
+  inherit (den.lib.stageTypes) stageTreeType;
 
   namespaceType = lib.types.submodule (
     nsArgs@{ name, ... }:
-    let
-      nsCtxApply = ctxApply nsArgs.config.ctx;
-      inherit (den.lib.ctxTypes nsCtxApply) ctxTreeType;
-    in
     {
-      options.ctx = lib.mkOption {
```
vic, line 13: Since we are keeping `den.ctx` compatibility, maybe we should add an alias here, otherwise people using `ns.ctx.x` are defining an aspect due to the freeform.

nix/lib/home-env.nix
```diff
@@ -49,21 +56,14 @@ let
...
-  intoClassUsers =
-    className:
-    { host }:
-    map (user: { inherit host user; }) (
-      lib.filter (u: lib.elem className u.classes) (lib.attrValues host.users)
-    );
-
   userEnvAspect =
     ctxName:
     { host, user }:
```
vic: Q: Since these args are context via handlers, can we change this to be: `{ host, user, class }:` right?

sini: > `host` and `user` come from ctx handlers (constantHandler), but `class` comes from a different path — the tree handler's class collector (`emit-class` effect). The double function reflects two-phase resolution: first entity context via handlers, then aspect-level class dispatch via tree traversal. 
> 
>  Flattening to `{ host, user, class, ... }:` would require `class` to also be resolvable as a ctx handler, which it currently isn't — it's injected by `classCollectorHandler` during `aspectToEffect` traversal, not by `constantHandler`. 


Edit: Technically correct method, incorrect conclusion. Bind does not care about the provenance of a param.

nix/lib/namespace.nix
sini: Agreed that `__ctx` is redundant — it's always a snapshot of what went into `constantHandler` to build `__scopeHandlers`. Every read site uses `or {}` fallback.

Consolidating to `__scopeHandlers` only makes sense. The 5 read sites that use `__ctx` would need to extract values from handlers instead (inverse of constantHandler), or we add a small helper. Tests in checkmate that assert on `__ctx.host` etc would need updating.

This is a good followup — want me to spike it on this branch or save it for a dedicated cleanup?

nix/lib/namespace-types.nix
sini: Good catch. Since `options.ctx` was removed from namespaces (commit 5aaadf44), any `ns.ctx.x` falls through the freeform `aspectsType` and silently creates an aspect entry — definitely not what users expect.

We should add a ctx alias in namespace-types.nix that forwards to stages, matching what the global compat shim does. Something like:

```nix
options.ctx = lib.mkOption {
  visible = false;
  type = lib.types.lazyAttrsOf stageTreeType;
};
config.stages = lib.mkAliasDefinitions options.ctx;
```

This way `ns.ctx.x` forwards to `ns.stages.x` with the same deprecation pattern.

nix/lib/namespace.nix
vic: on this branch, please.

sini: Inbound now.

nix/lib/namespace-types.nix
sini: It'll be in the next commit.

nix/lib/home-env.nix
vic: I was expecting `class` like any other param to be contextual. To me all params in Den come from context, not manually threading args.


As a side-note: I remember we had a `aspect-chain` being contextual (fixed to [self]) but don't know about now.

nix/lib/namespace.nix
vic: Also you can merge your other two PRs here, since they depend on this one, I prefer to have a better picture of what would be merged on main.

I mean, without those PRs already here this would not get merged.

vic: Stacking is good, as long as base PRs can be merged cleanly without issues.

sini:
@vic asked to merge the policy changes directly into this branch, so I have done so.

nix/lib/home-env.nix
sini: You're right — I was wrong in my earlier reply. `class` IS in constantHandler (pipeline.nix:56 `inherit class;`), so `fx.bind.fn` can resolve it alongside host/user. The double function is unnecessary.

`aspect-chain` is also there (pipeline.nix:57), fixed to `[ ]` at root (was `[ self ]` — looks like it changed). Both are pipeline-internal constants available to all parametric functions.

Will flatten `{ host, user }: { class, ... }:` → `{ host, user, class, ... }:` in home-env.nix.

sini: Sorry for letting Claude mislead you. I got distracted addressing your other two feedback and it was very eager in it's dispatch. :)

sini: I can assure you it actually behaves like you feel it should.

sini:
Added docs revamp

sini:
Added diagrams -- need a bit more fine tuning, but I think they'll be ready to ship before @vic actually finishes review so merging them in. :)

docs/src/content/docs/explanation/context-pipeline.mdx
```diff
@@ -1,99 +1,120 @@
...
+     homestage --> hmc["homeConfigurations.alice"]
+   ```
 
-	```mermaid
-	flowchart TD
-	  home["den.homes.x86_64-linux.alice"] --> homectx["den.ctx.home {home}"]
-	  homectx --> resolve["fixedTo {home} aspects.alice"]
-	  resolve --> hmc["homeConfigurations.alice"]
-	```
+   Home stages have no `host` in context, so policies and provides requiring
```
theutz: Does this HAVE to be the case? I understand that this is how things are now, but it’d be pretty neat if I could use my host schema data sometimes when I’m defining a standalone home if the home includes the hostname.

sini: Technically you could add a policy that would provide it.

theutz: That would be fantastic. Part of the reason I use standalone homes is that the build times for Nix Darwin can often be dreadfully slow. 30-50s to eval, then if you use Homebrew in nix-darwin, you have to wait for Brew to update, and THEN it runs your Home Manager config. So, like 1-2 minutes sometimes just to change an option on your tmux.

Anyways, all that to say, I very rarely have any actual `config` values that I need to share between those domains. It's usually more of the metadata kinda stuff you'd put in schemas.

So, yeah, I'll be building a policy like that and adding it to den.ful. :D

theutz:
Updated to the latest version of this, ran it again, and everything is still working. :)

sini:
Rendered docs can be viewed here: https://den.json64.dev/

sini:
@vic I rolled this spec in, since it's a compelling story that aligns with the other work: [parametric class application](https://gist.github.com/sini/be1bce0e9d54f57f3040a60648fedbef)

I will defer [traits/capabilities](https://gist.github.com/sini/ef23f526a49475932e74fc3baa21f6a5) as followup work.

docs/src/content/docs/explanation/context-pipeline.mdx
vic: > Home stages have no `host` in context

I believe they DO have it. IIRC correctly, the context for homes is built by filtering the entity attrs by `[ "host" "home" "user" ]` and it turnsout homes do have host and user attrs, so they might be in context already. 

NOTE: I remember reading that in the code, but I'm trying to find it, it is where entity.resolved is defined, IIRC

vic: `homeType` has:

```nix
        config._module.args.home = config;
        config._module.args.host = hostByName;
        config._module.args.user = userByName;
```

and 

options.nix does:

```nix
 ctx = lib.filterAttrs (n: v: isContextArg n && v != null) config._module.args 
```

where isContextArg includes `den.schema.<name>` or `den.stages.<name>`

```nix
  knownKinds = builtins.attrNames (den.stages or { });
```

that means for a home, you have { home, host, user } in context.

vic: Basically: If an `entity._module.args.<name>` has a corresponding `den.stages.<name>` or `den.schema.<name>`, you will have that `<name>` in context.

sini: It'll parametric-ly expand for any user defined kind, so long as there's a policy exposing it into the pipeline.

Home's being standalone I don't think have a default policy exposing host, but they could use home.name, .splitAt @, maybeGet den.hosts.<split:2>?

sini:
Vic has asked me to implement and roll in the traits spec... sooo.... here's to one mega-commit. :)

vic:
> Ok, I'd say we merge json(sini)  branch, i just need two or three more people help me by going quickly thu the diff, not trying to make sense of it all (i will resist my brain from doing that), just seeing there's nothing strange from the agent's code, I know json himself has already made some AI self-review passes on the code. If you do, please add comments to the PR or approve it.
>
> from @vic [at matrix](https://matrix.to/#/!mafjNSATIwbTXcmNzX:matrix.org/$1NJQGv9TolwLszUoEI2TO3tR6ceDbgqgW2D6tXvj_Do?via=matrix.org&via=rattlesnake.fr&via=crtified.me
)

drupol:
I love this, I'm looking for the merge now :)

docs/src/content/docs/explanation/diagrams.mdx
```diff
@@ -0,0 +1,142 @@
...
+
+import { Aside } from '@astrojs/starlight/components';
+
+<Aside title="Source" icon="github">
+[`nix/lib/diag/`](https://github.com/denful/den/tree/main/nix/lib/diag)
+</Aside>
+
+## What it does
+
+The `diag` library lets you visualize how Den resolves aspects -- which aspects include which, how policies fan out, which classes each aspect contributes to. It transforms structured trace data from the effects pipeline into a format-agnostic graph IR, applies filters and reshapes, then renders into Mermaid, GraphViz DOT, PlantUML, C4, and other formats.
```
drupol:
```suggestion
The `diag` library lets you visualize how Den resolves aspects -- which aspects include which, how policies fan out, which classes each aspect contributes to. It transforms structured trace data from the effects pipeline into a format-agnostic graph Intermediate Representation (Graph IR), applies filters and reshapes, then renders into Mermaid, GraphViz DOT, PlantUML, C4, and other formats.
```

docs/src/components/Ad.astro
```diff
@@ -20,8 +20,8 @@ import { LinkButton, Icon } from '@astrojs/starlight/components';
       <li><a href="https://github.com/vic/flake-aspects">flake-aspects</a></li>
       <li><a href="https://github.com/vic/flake-file">flake-file</a></li>
       <li><a href="https://github.com/vic/with-inputs">with-inputs</a></li>
```
vic:
```suggestion
<li><a href="https://github.com/denful/flake-aspects">flake-aspects</a></li>
<li><a href="https://github.com/denful/flake-file">flake-file</a></li>
<li><a href="https://github.com/denful/with-inputs">with-inputs</a></li>
```

docs/src/components/Ad.astro
```diff
@@ -20,8 +20,8 @@ import { LinkButton, Icon } from '@astrojs/starlight/components';
...
-      <li><a href="https://github.com/vic/denful">denful</a></li>
-      <li><a href="https://github.com/vic/dendrix">dendrix</a></li>
+      <li><a href="https://github.com/denful/denful">denful</a></li>
+      <li><a href="https://github.com/denful/dendrix">dendrix</a></li>
     </ul>
 
     <LinkButton icon="github" href="https://github.com/vic?tab=repositories" variant="minimal"><code>More <strong>@vic</strong> OpenSource</code></LinkButton>
```
vic:
```suggestion
    <LinkButton icon="github" href="https://github.com/denful?tab=repositories" variant="minimal"><code>More <strong>@denful</strong> OpenSource</code></LinkButton>
```

docs/src/content/docs/explanation/coming-from.mdx
```diff
@@ -0,0 +1,112 @@
...
+## Coming from flake-parts
+
+| flake-parts concept | Den equivalent |
+|---|---|
+| `perSystem` | `den.stages.flake-system` -- flake-level [policies](/explanation/policies/) fan out per system |
+| `flake.nixosConfigurations` | `den.hosts` -- Den generates flake outputs from entity declarations |
+| `mkFlake` | `inputs.den.flakeModule` or `inputs.den.denModule` -- Den is a flake-parts module |
+| `flakeModules` for sharing | `den.ful` namespaces -- shareable aspect libraries |
+
+Den is built on flake-parts, so the two are complementary. You can use `perSystem`
```
vic: > Den is built on flake-parts

This gives the impression that Den depends on flake-parts. Remove this part, I seem to remember having read this in other doc page.

docs/src/content/docs/explanation/core-principles.mdx
```diff
@@ -5,9 +5,9 @@ description: The two design principles behind Den.
 
 import { Aside } from '@astrojs/starlight/components';
 
-**Aspects** and **Contexts** are Den core arquitectural principles.
+**Aspects** and **Contexts** are Den core architectural principles.
 
-It is how Den supports different platform configurations, packages/checks flake outputs, different home environments suppport,
```
vic, line 10: `suppport` :D

docs/src/content/docs/explanation/core-principles.mdx
```diff
@@ -64,65 +64,72 @@ Nix classes are NOT an artificial concept created by Den. There are [many nix cl
 
 Most importantly, the context `{host,user}` here are **not** `_module.args` **nor** `specialArgs`, it is an actual function, not a module-looking-as-a-function. This means config can depend on context without Nix infinite loops.
 
+<Aside type="caution" title="Den context ≠ NixOS module args">
+When Den docs say "context" they mean the pipeline data shape (`{ host }`, `{ host, user }`,
```
vic: Doc is using now two different ways to talk about these pipeline params, in some we "context" (terminology we have always used) and some other say "den params". I'd prefer if we decide on only one.

docs/src/content/docs/explanation/entities.mdx
```diff
@@ -0,0 +1,149 @@
...
+
+- **Users** — declared inline on a host via `den.hosts.<sys>.<host>.users.<name>`.
+  Users are not top-level entities; they live within their host.
+
+```nix
+{
+  den.hosts.x86_64-linux.igloo = {
+    users.alice = { };
+    users.bob = { };
+  };
```
vic:
```suggestion
  den.hosts.x86_64-linux.igloo.users = {
    alice = { };
    bob = { };
  };
```

docs/src/content/docs/explanation/entities.mdx
```diff
@@ -0,0 +1,149 @@
...
+
+  den.homes.aarch64-darwin.alice = { };
+}
+```
+
+Every entity carries a `class` that determines how it is instantiated and where its
+output lands. For hosts this is `"nixos"` (auto-detected from `x86_64-linux` / `aarch64-linux`)
+or `"darwin"` (from `*-darwin`); for homes it is `"homeManager"`. Override with
```
vic, line 49: Add: Users have a `classes` list declaring each home-environment class they want to use.

docs/src/content/docs/explanation/entities.mdx
```diff
@@ -0,0 +1,149 @@
...
+    options.copyright = lib.mkOption { default = "Copy-Left"; };
+  };
+}
+```
+
+Schema entries are `deferredModule` values — they can be plain attrsets with
+`options`/`config` or functions that receive the entity's module arguments.
+
+<Aside>
+`den.schema` uses a freeform type, so you can define custom kinds beyond the
```
vic, line 101: Maybe this is a good place to talk about `strict` mode ?

docs/src/content/docs/explanation/entities.mdx
```diff
@@ -0,0 +1,149 @@
...
+  den.schema.host.policies = [ "host-to-peers" ];
+
+  # Activate a policy for a specific host
+  den.hosts.x86_64-linux.igloo.policies = [ "microvm-guests" ];
+}
+```
+
+See [Policies](/explanation/policies/) for how activation is used during resolution.
+
+## Freeform attributes
```
vic, line 137: Also talk about strict mode here.

docs/src/content/docs/explanation/library-vs-framework.mdx
```diff
@@ -11,13 +11,13 @@ Den's core (`/default.nix`) is domain-agnostic. It provides:
...
-| `parametric.exactly` | Match functions with exactly the given params |
+| `parametric` | *(deprecated)* Context dispatch now happens automatically |
+| `parametric.atLeast` | *(deprecated)* Use bare functions — dispatch is automatic |
+| `parametric.exactly` | *(deprecated)* Use bare functions — dispatch is automatic |
 | `parametric.fixedTo` | Apply an aspect with a fixed context |
 | `parametric.expands` | Extend context before dispatch |
 | `canTake` | Check if a function accepts given arguments |
-| `take.atLeast` / `take.exactly` | Conditionally apply functions |
+| `take.atLeast` / `take.exactly` | *(deprecated)* Dispatch is now handled by the pipeline |
 | `statics` | Extract only static includes from an aspect |
```
vic: `statics` was removed!

docs/src/content/docs/explanation/library-vs-framework.mdx
```diff
@@ -59,7 +60,7 @@ module = den.lib.aspects.resolve "terranix" [] aspect;
 On top of the library, Den provides `modules/` which implement:
 
 - **Schema types** (`den.hosts`, `den.homes`) for declaring NixOS/Darwin/HM entities
-- **Context pipeline** (`den.ctx.host`, `den.ctx.user`, `den.ctx.home`) for OS configurations
+- **Stages & Policies** (`den.stages`, `den.policies`) for the resolution pipeline
```
vic: Reminder that Stages are going away in favor of using aspects only.

docs/src/content/docs/explanation/class-modules.mdx
```diff
@@ -0,0 +1,171 @@
...
+{
+  nixos = { host, user, config, ... }: {
+    users.users.${user.name}.description = "${host.name}/${user.name}";
+  };
+}
+```
+
+### Full application
+
+When **all** function args are den args (no module-system args at all), the function is called directly and its return value becomes the class module. This handles the pattern where a function returns another function:
```
vic: Here we use `den args` other places use `Parametric context args` let us use only one terminology on all docs.

docs/src/content/docs/explanation/class-modules.mdx
```diff
@@ -0,0 +1,171 @@
...
+}
+```
+
+The pipeline's `wrapClassModule` intercepts class modules before they reach the OS module system. It inspects `builtins.functionArgs`, identifies which args exist in the pipeline context (`ctx`), pre-applies those values, and advertises only the remaining args to the module system via `lib.setFunctionArgs`. The module never knows it was wrapped.
+
+## How It Works
+
+1. `builtins.functionArgs module` returns the full set of declared args.
+2. Each arg name is checked against the pipeline context: a key is a den arg if `ctx ? ${key}`.
+3. Den args are pre-applied. The remaining args are advertised to the module system.
```
vic: `Den args` -> `Context args` ?

docs/src/content/docs/explanation/class-modules.mdx
```diff
@@ -0,0 +1,171 @@
...
+
+import { Aside } from '@astrojs/starlight/components';
+
+<Aside title="Source" icon="github">
+[`nix/lib/aspects/fx/aspect.nix`](https://github.com/denful/den/blob/main/nix/lib/aspects/fx/aspect.nix)
+</Aside>
+
+## The Problem
+
+Accessing den context (like `host` or `user`) inside a class module previously required a two-layer curried function. The outer function captured den args via parametric dispatch, and the inner function was the actual NixOS/Darwin/home-manager module:
```
vic: There's nothing wrong with using functions `{ host }: { nixos = { pkgs }: ... }` they are not curried, remove this line, it gives the impression that this form is not recommended when it is. 

Do not mention it as Problem/Solution, access to ctx args on classModules is a **convenience**.

docs/src/content/docs/explanation/class-modules.mdx
```diff
@@ -0,0 +1,171 @@
...
+```nix
+# Flat form -- den args and module-system args side by side
+{
+  nixos = { host, config, pkgs, ... }: {
+    networking.hostName = host.name;
+  };
+}
+```
+
+The pipeline's `wrapClassModule` intercepts class modules before they reach the OS module system. It inspects `builtins.functionArgs`, identifies which args exist in the pipeline context (`ctx`), pre-applies those values, and advertises only the remaining args to the module system via `lib.setFunctionArgs`. The module never knows it was wrapped.
```
vic: Not sure if we want to leak implementation details like `wrapClassModule`, but it is ok if we talk about using `lib.setFunctionArgs` to show how magic happens at module-level eval.

docs/src/content/docs/explanation/class-modules.mdx
```diff
@@ -0,0 +1,171 @@
...
+
+A static aspect whose `nixos` class module requests `host` directly:
+
+```nix
+{
+  den.aspects.my-aspect = {
+    nixos = { host, config, ... }: {
+      networking.hostName = host.name;
+    };
+  };
```
vic, line 57: Add a `homeManager = { user, config, ... }: ... ` example here. Both `home` and `user` come from context.

docs/src/content/docs/explanation/class-modules.mdx
```diff
@@ -0,0 +1,171 @@
...
+[`nix/lib/aspects/fx/aspect.nix`](https://github.com/denful/den/blob/main/nix/lib/aspects/fx/aspect.nix)
+</Aside>
+
+## The Problem
+
+Accessing den context (like `host` or `user`) inside a class module previously required a two-layer curried function. The outer function captured den args via parametric dispatch, and the inner function was the actual NixOS/Darwin/home-manager module:
+
+```nix
+# Two-layer form (still works, but verbose)
+{ host, ... }: {
```
vic: convention, for aspect-level context args, we don't need ellipsis anymore, they don't make sense, we used ellipsis because how our old context-dispatch worked. There's no way a context-level-function takes more args than the ones it defines since they come from context directly.

So I'd recommend in all docs and code examples, we do not use ellipsis on context-level args. but only on classModule args.

docs/src/content/docs/explanation/class-modules.mdx
```diff
@@ -0,0 +1,171 @@
...
+
+```nix
+{
+  nixos = { host }: { config, ... }: {
+    networking.hostName = host.name;
+  };
+}
+```
+
+Here `{ host }` has no `...` and no module-system args. `wrapClassModule` calls it with `{ host = ctx.host; }` and uses the returned function as the class module.
```
vic: Again avoid mentioning implementation details like `wrapClassModule` unless they are key for people understanding.

docs/src/content/docs/explanation/class-modules.mdx
```diff
@@ -0,0 +1,171 @@
...
+
+<Aside type="caution">
+Flat-form class modules **must** include `...` in their argument pattern.
+</Aside>
+
+The NixOS module system passes extra args (`config`, `options`, `lib`, `pkgs`, `modulesPath`, and anything from `specialArgs` or `_module.args`) to every module function. Without `...`, the module system's extra args cause an eval error. This is already conventional for NixOS modules, but worth noting since the flat form makes it easy to forget.
+
+The exception is **full application** -- if every arg is a den arg (e.g., `{ host }:`), the function is called directly by den and never reaches the module system, so `...` is not needed.
+
+## Deprecated Helpers
```
vic: remove this section about `den.lib.*` and migration from it. is that worth mentioning here? I believe we have a ctx migration page, link to that.

docs/src/content/docs/explanation/class-modules.mdx
```diff
@@ -0,0 +1,171 @@
...
+
+Migration is straightforward: move the den args from the outer function into the class module's arg pattern and add `...` if not already present.
+
+## Collision Policies
+
+A collision occurs when den wants to inject an arg (e.g., `host`) but the module system also provides a value for the same name -- via `specialArgs` or `_module.args`. By default, this throws an error to force an explicit resolution.
+
+### Resolution levels
+
+Policy is resolved from three levels (first match wins):
```
vic: I know maybe it is my mind, but using Policy here makes me suddenly think we are talking about `den.policies` here, use another word.

docs/src/content/docs/explanation/core-principles.mdx
vic: >  data shape

We used that term before because each context stage in previous pipeline had a fixed data shape. However that is no longer the case,

`{ host, user }:` is the same as `{ host }: { user }`, and you can use `{ user }` only if you want, or even `{ user, class }` anything that comes from context is a param, so there's no "context shape" anymore.

docs/src/content/docs/explanation/entities.mdx
```diff
@@ -0,0 +1,149 @@
...
+by default `den.aspects.<name>`.
+
+<Aside type="tip" title="Class vs Entity Kind">
+Don't confuse **class** (`nixos`, `darwin`, `homeManager` — a Nix module evaluation
+domain) with **entity kind** (`host`, `user`, `home` — a Den data type). Classes
+determine _which module system_ evaluates the config. Entity kinds determine _how
+policies dispatch_.
+</Aside>
+
+## Schema: shared options across a kind
```
vic, line 74: Maybe add an aside/section mentioning:

An entity is like an "instance", a host can be seen as an instance of den.schema.host 

The "mixin" modules: 
A host entity has attrs from  
- itself: `{ config, ... }: { options.myOwnAttr }` (like meta-class in ruby)
- den.schema.host mixin
- den.schema.conf mixin
- hostType defined by Den


I'm not sure if it makes sense the meta-class/mixin analogies here. but they could be useful for people to get it.

vic (dismissed):

docs/src/content/docs/explanation/library-vs-framework.mdx
```diff
@@ -11,13 +11,13 @@ Den's core (`/default.nix`) is domain-agnostic. It provides:
...
+| `parametric.atLeast` | *(deprecated)* Use bare functions — dispatch is automatic |
+| `parametric.exactly` | *(deprecated)* Use bare functions — dispatch is automatic |
 | `parametric.fixedTo` | Apply an aspect with a fixed context |
 | `parametric.expands` | Extend context before dispatch |
 | `canTake` | Check if a function accepts given arguments |
-| `take.atLeast` / `take.exactly` | Conditionally apply functions |
+| `take.atLeast` / `take.exactly` | *(deprecated)* Dispatch is now handled by the pipeline |
 | `statics` | Extract only static includes from an aspect |
 | `owned` | Extract only owned configs (no includes, no functor) |
 | `isFn` | Check if value is a function or has `__functor` |
```
vic: remove isFn, I believe that is not even needed in code, since `lib.isFunction` exist, or is it for `__fn`, `__args` ? even so they are private APIs. remove anything we don't want people to us directly.

docs/src/content/docs/explanation/policies.mdx
```diff
@@ -0,0 +1,153 @@
...
+
+import { Aside } from '@astrojs/starlight/components';
+
+<Aside title="Source" icon="github">
+[`nix/lib/policy-types.nix`](https://github.com/denful/den/blob/main/nix/lib/policy-types.nix) --
+[`modules/policies/core.nix`](https://github.com/denful/den/blob/main/modules/policies/core.nix) --
+[`nix/lib/policy-inspect.nix`](https://github.com/denful/den/blob/main/nix/lib/policy-inspect.nix)
+</Aside>
+
+Policies are the second of Den's four concerns.
```
vic: `three concerns` when we remove stages, right ?

docs/src/content/docs/reference/aspects.mdx
```diff
@@ -69,38 +69,16 @@ An aspect is an attribute set with:
 Functions in `includes` receiving `{ class, aspect-chain }` are **static** --
 evaluated once during aspect resolution. Functions receiving context
 arguments (`{ host }`, `{ user }`, etc.) are **parametric** -- evaluated
-per context during `ctxApply`.
+per context during pipeline resolution.
 
 ### `meta.adapter`
```
vic: Does it make sense to keep meta.adapter since we have handleWith ?

docs/src/content/docs/explanation/core-principles.mdx
sini: arquitectural

sjcobb2022:
Currently on a6bd47c8 of #475 (latest) and all seems to be functioning.

https://github.com/sjcobb2022/nix-den

sini:
FYI, while all templates pass, I have introduced a major real-world regression that I'm actively working on. Rather than rolling back, I'm rolling forward. If you update and it's broken, just uh... don't update until I give the all-clear. :)

sini:
As far as I'm aware, that fixes the immediate regressions.

drupol:
Hello there,

I tested this against `drupol/infra` this morning and got this: https://gist.github.com/drupol/eefd015f7b43212d4d61881984498a07

sini:
> Hello there,
> 
> I tested this against `drupol/infra` this morning and got this: https://gist.github.com/drupol/eefd015f7b43212d4d61881984498a07

Thank you, I'm going to try and strip your infra-private flake out and reproduce. The trace data is useful. :)

sini:
Let me know if there are any new regressions -- this was a pretty big refactor. I tested with @theutz and @drupol's configs and things looked good at first glance but I didn't delve deeper than just 'system eval produces derivation'.

drupol:
> Let me know if there are any new regressions -- this was a pretty big refactor. I tested with @theutz and @drupol's configs and things looked good at first glance but I didn't delve deeper than just 'system eval produces derivation'.

Everything works perfectly, thank you !

theutz (dismissed):

theutz (approved):

drupol (approved):

drupol:
Massive :) congratulations!

PR #477, opened 2026-04-24
feat: policy foundations — as field, activation, provide-to, inspect

sini:
## Summary

- **Policy `as` field** — context key aliasing for sibling routing (e.g., host→host uses `as = "peer"` to avoid key collision)
- **4-level activation model** — core (`_core`), global (`den.default.policies`), schema-kind and entity-instance (`entity.policies` via `schema.conf` pattern)
- **provide-to structural key + collection handler** — cross-entity data routing infrastructure; aspects declare `provide-to.label = data`, handler accumulates in pipeline state
- **Policy inspect utility** — `den.lib.policyInspect.inspect` for lightweight debugging without full pipeline execution

## Test plan

- [x] 512/512 CI tests pass (`nix develop -c just ci`)
- [x] 20 new tests across 4 test files covering all features
- [x] Activation model: core always-active, default opt-in, schema-kind scoped, entity-instance scoped, non-activated excluded
- [x] provide-to: collection in state, emission shape, multiple labels, structural key not treated as class
- [x] Inspect: core policies, resolved targets, activation-aware, sibling routing detection

### DIFF
--- a/modules/aspects/provides/hjem.nix
+++ b/modules/aspects/provides/hjem.nix
@@ -41,4 +41,9 @@ in
       resolve = mkIntoClassUsers "hjem";
     };
   };
+
+  den.schema.host.policies = [
+    "host-to-hjem-host"
+    "hjem-host-to-hjem-user"
+  ];
 }
--- a/modules/aspects/provides/home-manager.nix
+++ b/modules/aspects/provides/home-manager.nix
@@ -44,4 +44,9 @@ in
       resolve = mkIntoClassUsers "homeManager";
     };
   };
+
+  den.schema.host.policies = [
+    "host-to-hm-host"
+    "hm-host-to-hm-user"
+  ];
 }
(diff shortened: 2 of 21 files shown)


modules/aspects/provides/hjem.nix
```diff
@@ -41,4 +41,9 @@ in
       resolve = mkIntoClassUsers "hjem";
     };
   };
+
+  den.schema.host.policies = [
```
vic, line 45: these policies should be part of the result of home-env.

vic (changes requested):

modules/aspects/provides/hjem.nix
sini: Addressed in the fx-pipeline branch, will close this PR out.

sini:
We'll merge this directly in the feat/fx-pipeline branch

PR #478, opened 2026-04-24
feat: per-policy dispatch, routing, provide-to distribution

sini:
## Summary

- **Per-policy named effect dispatch** — each policy compiles to a `policy:<name>` handler, enabling granular tracing and individual override via `scope.provide`. Activation model enforced end-to-end.
- **Routing decision** — sibling targets (`policy.from == policy.to`) route through provide-to; child targets resolve locally
- **Distribution library** — `groupByTarget`, `mkProvideToHandlers` for phase 2 cross-entity data injection
- **Forward propagation** — sub-pipeline provide-to emissions splice into parent's thunk chain, deferring list evaluation through the deepSeq barrier
- **Test policy activation** — all custom test policies activated via `den.default.policies`; battery policies at schema-kind level

Builds on #477 (policy foundations).

## Test plan

- [ ] 513/513 CI tests pass (`nix develop -c just ci`)
- [ ] Per-policy traces visible for each fired policy
- [ ] Sibling routing test: host-to-peer emits provide-to, not local resolution
- [ ] Forward propagation: sub-pipeline provide-to accessible in parent state
- [ ] No double-dispatch: `mergePolicyInto` removed from `resolveStage`
- [ ] Same-path policy merging: multiple policies targeting same stage accumulate contexts

### DIFF
--- a/nix/lib/aspects/default.nix
+++ b/nix/lib/aspects/default.nix
@@ -62,11 +62,24 @@ let
       self = wrapped;
     };
 
+  # Like resolve but returns full pipeline result including state.provideTo.
+  fxResolveTreeFull =
+    class: resolved:
+    let
+      wrapped = normalizeRoot resolved;
+      ctx = resolved.__ctx or { };
+    in
+    fx.pipeline.fxFullResolve {
+      inherit class ctx;
+      self = wrapped;
+    };
+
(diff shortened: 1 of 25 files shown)


sini:
We'll merge this directly in the feat/fx-pipeline branch

PR #485, opened 2026-05-08
feat: default command and args

theutz:
Make the args handling a bit more flexible.


theutz:
The docs build error seems to be related to a node module that I don't know how to fix and don't think is relevant to my minor changes

sini (approved):

PR #495, opened 2026-05-09
fix: unfree provider for non-module-system user classes

sini:
## Summary
- When a user has `classes = ["user"]` (no homeManager), `den.provides.unfree` resolved with `class = "user"` — not in `validClasses` — silently dropping the `unfree.packages` emission entirely
- Fix: extend the `hostModule` condition to emit to the host's OS class for any non-module-system class, not just `homeManager`
- All existing unfree tests pass; verified with a reproduction case matching the reported config

## Test plan
- [x] Existing `unfree` test suite passes (test-packages-set-on-nixos, test-packages-set-on-home-manager, test-user-class-works)
- [x] Reproduced: user aspect chain (user → development → ollama → unfree) with `classes = ["user"]` now correctly populates `igloo.unfree.packages`

### DIFF
--- a/modules/aspects/provides/unfree/unfree.nix
+++ b/modules/aspects/provides/unfree/unfree.nix
@@ -31,11 +31,17 @@ let
         ];
         classModule =
           if builtins.elem class validClasses then { ${class}.unfree.packages = allowed-names; } else { };
-        # When resolving for homeManager at user scope, also emit to the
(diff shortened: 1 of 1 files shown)


theutz (approved):

drupol:
The insecure battery most probably suffer from the same issue ?

sini:
> The insecure battery most probably suffer from the same issue ?

Yes, I was going to do a full audit after breakfast.

PR #497, opened 2026-05-09
fix: insecure provider for non-module-system user classes

sini:
## Summary
- Same bug as #495 (unfree): when a user has `classes = ["user"]` (no homeManager), `den.provides.insecure` resolved with `class = "user"` — not in `validClasses` — silently dropping the `permittedInsecurePackages` emission
- Fix: add `host` arg and emit to the host's OS class for any non-module-system class, matching the unfree provider fix
- Audited all other providers in `modules/aspects/provides/` — none are affected

## Test plan
- [x] Existing `insecure` test suite passes
- [x] Verified: user aspect with `classes = ["user"]` now correctly populates `igloo.permittedInsecurePackages.packages`

### DIFF
--- a/modules/aspects/provides/insecure/insecure.nix
+++ b/modules/aspects/provides/insecure/insecure.nix
@@ -18,21 +18,39 @@ let
       "provides"
     ];
     __fn =
-      { class, ... }:
-      if
-        (builtins.elem class [
+      {
+        class,
+        host ? null,
+        ...
+      }:
+      let
+        validClasses = [
           "nixos"
           "darwin"
           "homeManager"
-        ])
-      then
-        {
-          ${class}.permittedInsecurePackages.packages = allowed-names;
-        }
-      else
-        { };
+        ];
+        classModule =
+          if builtins.elem class validClasses then
+            { ${class}.permittedInsecurePackages.packages = allowed-names; }
+          else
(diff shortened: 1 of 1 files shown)


drupol (approved):

theutz (approved):

modules/aspects/provides/insecure/insecure.nix
```diff
@@ -18,21 +18,39 @@ let
...
+          if builtins.elem class validClasses then
+            { ${class}.permittedInsecurePackages.packages = allowed-names; }
+          else
+            { };
+        # When resolving for homeManager or a non-module-system class (e.g.
+        # "user"), also emit to the host's OS class so
+        # nixpkgs.config.permittedInsecurePackages covers these packages.
+        hostModule =
+          if
+            (class == "homeManager" || !builtins.elem class validClasses)
```
drupol, line 42: Stupid question but: is there a way to avoid hardcoding the class here ?

sini: I mean, it's a feature specific to homeManager (optionally) having it's own internal package registry/store, so it's not really 'hard coded' as it is home manager specific.

drupol (approved):

PR #499, opened 2026-05-09
fix: policy.for accepts list of entities and hardens id_hash matching

sini:
## Summary
- `policy.for` now accepts a single entity or a list of entities as its first argument — matching fires if the context entity's `id_hash` matches any entity in the list
- Fixes `policy.for` never applying when users pass `policy.for [ host1 host2 ] (policy.include ...)` — the list had no `id_hash` attribute, so the comparison always failed
- Hardens `id_hash` to require non-null on both sides (prevents `null == null` false positives) and filters `_`-prefixed options from the identity hash computation

## Test plan
- [x] 760/760 CI tests pass
- [x] New test: `test-for-entity-list` — list of entities matches any member
- [x] New test: `test-for-mixed-hash-list` — list with mixed `id_hash` presence, hash-less entries silently dropped
- [x] New test: `test-for-no-id-hash` — entities without `id_hash` never match
- [ ] Manual verification with the reported user pattern: `policy.for [ x86_64-linux.yalova ] (policy.include aspects.tlp)`

### DIFF
--- a/modules/options.nix
+++ b/modules/options.nix
@@ -102,8 +102,11 @@ let
                 default =
                   let
                     isPrimitive =
-                      _: opt:
-                      (opt ? type) && builtins.elem (opt.type.name or "") primitiveTypeNames && !(opt.internal or false);
+                      name: opt:
+                      !(lib.hasPrefix "_" name)
+                      && (opt ? type)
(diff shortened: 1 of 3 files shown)


drupol (approved):
Diff LGTM

PR #501, opened 2026-05-10
fix: nested classification, hm-host bridge, bare aspect access

sini:
## Summary

Three post-fx-pipeline regressions reported by a user after updating:

- **Nested classification false positive** — freeform data keys with sub-keys matching registered class names (e.g. `git.user` colliding with `den.classes.user`) were misclassified as nested aspects, causing content to leak into class routing (`users.users.loom.email` does not exist). Added `looksLikeClassContent` guard to `hasRecognizedSubKeys`.

- **`den.schema.hm-host` not consumed** — the ctx-shim forwarded `den.ctx.hm-host` to `den.schema.hm-host.includes`, but nothing in the pipeline read that schema. Added a `policy.when`-guarded bridge in the HM battery that injects `hm-host` schema includes via `policy.include` when the host has homeManager users.

- **Bare nested aspect access** — `aspectContentType` wrapped all freeform values into opaque `__contentValues` wrappers, preventing direct attribute access like `gloom.apps.polybar.razermon`. Shallow-merge attrset definition values onto the wrapper so sub-keys are accessible for config access and includes.

## Test plan

- [x] 758/758 CI tests pass (`nix develop -c just ci`)
- [x] New regression test `test-class-name-collision-not-nested`
- [x] User's dotfiles: all 8 nixosConfigurations eval on both `main` and `new_den` (migration) branches
- [x] `den.ctx.hm-host` deprecation warning fires and `useGlobalPkgs` propagates correctly

### DIFF
--- a/modules/aspects/provides/home-manager.nix
+++ b/modules/aspects/provides/home-manager.nix
@@ -1,10 +1,12 @@
 {
   den,
+  config,
+  lib,
   inputs,
   ...
 }:
 let
-  inherit (den.lib.home-env) makeHomeEnv;
+  inherit (den.lib.home-env) makeHomeEnv mkDetectHost;
 
   result = makeHomeEnv {
     className = "homeManager";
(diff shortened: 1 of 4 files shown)


modules/aspects/provides/home-manager.nix
```diff
@@ -20,10 +22,30 @@ let
...
+  );
+
 in
 {
   den.schema.host.imports = [ result.hostConf ];
-  den.schema.host.includes = [ result.battery ];
+  den.schema.host.includes = [
+    result.battery
+  ]
+  ++ lib.optional (hmHostSchemaIncludes != [ ]) {
```
drupol: Nit: using optionals might be more explicit here. WDYT ?

sini: k

drupol (approved):

PR #503, opened 2026-05-10
feat: add pipe.as stage for cross-quirk adaptation

sini:
## Summary

- Adds `pipe.as` stage that renames pipe output so data from one quirk can be delivered under a different quirk name
- Enables derived quirks — quirks with no native emitters, populated entirely via collect + transform + rename from other pipes
- Composes with all existing stages (`pipe.collect`, `pipe.transform`, `pipe.filter`, `pipe.to`)
- Self-targeting (`pipe.as "same-pipe"`) throws a clear error

## Changes

- `nix/lib/policy-effects.nix`: `pipe.as` stage constructor
- `nix/lib/aspects/fx/assemble-pipes.nix`: Assembly logic for untargeted and targeted `pipe.as` routing, `mkCombinedBase` helper extraction
- `templates/ci/modules/features/pipe-policy.nix`: 6 new tests (basic, transform, collect, pipe.to, no-emitter quirk, self-error)
- `docs/`: Updated explanation, guide, and reference docs for `pipe.as`

## Test plan

- [x] 767/767 CI tests pass
- [x] 6 new pipe-policy tests cover all `pipe.as` combinations
- [x] Existing pipe tests unaffected (no regressions)

### DIFF
--- a/docs/src/content/docs/explanation/quirks-and-pipes.mdx
+++ b/docs/src/content/docs/explanation/quirks-and-pipes.mdx
@@ -172,6 +172,25 @@ den.policies.expose-prefs = { host, user, ... }:
   [ (pipe.from "prefs" [ pipe.expose ]) ];
 ```
 
+### Renaming with `pipe.as`
+
+`pipe.as` delivers data under a different quirk name. This creates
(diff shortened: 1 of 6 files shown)


vic (approved):
nice!

PR #504, opened 2026-05-10
refactor: Move all batteries to `den.batteries` make `den.provides` and `den._` aliases.

vic:
Also removed `._.` and `.provides.` since they are no longer needed (kept templates/ci as is)


docs/src/content/docs/motivation.mdx
sini, line 1: You probably should apply your regex's in opposite order, den._.unfree -> den.batteries.unfree 

You can alias den._. to batteries for legacy, but I don't think we should document it.

drupol (commented):
Big yes for this!

docs/src/content/docs/motivation.mdx
vic: done, I reapplied sed (everywhere except templates/ci so we check that aliases actually work as people use them today) 

den._ -> den.batteries
den.provides -> den.batteries
`._.` -> `.`
`.provides.` -> `.`

sini (approved):
LGTM

PR #506, opened 2026-05-10
fix: forward provides children onto aspect root

sini:
## Summary

- Forward `provides` children onto the merged aspect root in `mergeWithAspectMeta`, so `aspect.docker` resolves to `aspect.provides.docker` without requiring the `_.docker` accessor
- Tag forwarded keys via `__providesForwarded` so the pipeline's key classification skips them — prevents double-processing alongside includes
- Add `__providesForwarded` to `structuralKeysSet`

Fixes regression from #504 which removed `_.docker` usage but did not implement the forwarding, breaking `den.aspects.virtualization.docker` access in diagram-demo template.

## Test plan

- [x] 771/771 CI tests pass
- [x] diagram-demo template evaluates (148 checks)
- [x] den-configs helix.enable = true
- [x] `aspect-path.test-excludeAspect-by-provider` passes (validates forwarded keys don't affect pipeline classification)

### DIFF
--- a/nix/lib/aspects/fx/key-classification.nix
+++ b/nix/lib/aspects/fx/key-classification.nix
@@ -25,6 +25,7 @@ let
     "__parametricResolvedArgs"
     "__contentValues"
     "__provider"
+    "__providesForwarded"
     "_module"
     "_"
   ] (_: true);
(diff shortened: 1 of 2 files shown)


vic:
omg thanks! and sorry for the regression.

vic (approved):

PR #507, opened 2026-05-11
fix: stable identity for bracket-resolved bare attrsets

sini:
## Summary

Bracket resolution through forwarded attrs on content wrappers (e.g. `<gloom/apps/polybar/razermon>`) produces bare attrsets without `__contentValues` or `__provider`. Without identity, these get anonymous names and dedup fails when the same content is reached via multiple resolution paths.

- **den-brackets.nix**: `tagProvider` injects `__provider` from the full bracket path onto bare attrset results
- **types.nix**: `providerType.merge` normalization detects `__provider` (not just `__contentValues`) and injects `name` + `meta.provider` for stable identity

Fixes module duplication where entries appeared twice — once from the direct include path and once from host-aspects re-resolution with anonymous identity.

## Test plan

- [x] 771/771 CI tests pass
- [x] den-configs polybar `modules-right` shows each module exactly once
- [x] diagram-demo template evaluates (148 checks)

### DIFF
--- a/nix/lib/aspects/types.nix
+++ b/nix/lib/aspects/types.nix
@@ -210,7 +210,11 @@ let
           # contain parametric functions.  Without this, the wrapper merges
           # through aspectSubmodule and the function is buried as a freeform
           # key.  Extract the function so existing dispatch handles it.
-          isContentWrapper = d: builtins.isAttrs d.value && d.value ? __contentValues && !(d.value ? __fn);
+          isContentWrapper =
+            d:
+            builtins.isAttrs d.value
+            && (d.value ? __contentValues || d.value ? __provider)
+            && !(d.value ? __fn);
           nameFromProvider =
             v:
             let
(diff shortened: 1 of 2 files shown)


nix/lib/aspects/types.nix
```diff
@@ -220,16 +224,20 @@ let
...
-                lib.isFunction cv.value
-                && (
-                  let
-                    args = builtins.functionArgs cv.value;
-                  in
-                  args != { } && !(args ? config) && !(args ? options)
-                )
-              ) d.value.__contentValues;
+              fns =
+                if d.value ? __contentValues then
```
drupol, line 228: Nit: Can't we use `lib.optionals` ? (to get rid of the `else [ ]`)

theutz (approved):
Bro

nix/lib/aspects/types.nix
sini: I gotchu next time I'm here. :D

PR #510, opened 2026-05-11
fix: guard home.class access in battery homeAspect functions

sini:
## Summary

- Guard `home.class` access in all battery `homeAspect` functions with conditional checks
- Prevents crash when `home` entity lacks a `class` attribute (e.g. no `den.homes` defined)
- Affects: unfree-predicate-builder, insecure-predicate-builder, flake-parts/self, flake-parts/inputs

## Test plan

- [x] 771/771 CI tests pass
- [x] unfree and standalone-homes suites pass

### DIFF
--- a/modules/aspects/batteries/flake-parts/inputs.nix
+++ b/modules/aspects/batteries/flake-parts/inputs.nix
@@ -1,4 +1,4 @@
-{ withSystem, ... }:
+{ lib, withSystem, ... }:
 let
   description = ''
     Provides the `flake-parts` `inputs'` (the flake's `inputs` with system pre-selected)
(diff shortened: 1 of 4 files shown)


modules/aspects/batteries/flake-parts/inputs.nix
```diff
@@ -56,7 +56,7 @@ let
     {
       name = "inputs'/home";
     }
-    // mkAspect home.class home.system;
+    // (if home ? class then mkAspect home.class home.system else { });
```
drupol:
```suggestion
    // (lib.optionalAttrs (home ? class) (mkAspect home.class home.system));
```

sini: had to import lib at the header too -- adopted

drupol (approved):

PR #512, opened 2026-05-11
fix: import-tree class resolution and _ alias on batteries

sini:
## Summary

- import-tree battery now scans `_<class>` directories eagerly via `builtins.readDir` instead of relying on the `class` scope parameter, which the fx-pipeline only provides once per scope (not once per class)
- Batteries with `__functor` (import-tree, forward, etc.) now forward `provides` children and the `_` alias in the `mergeFunctions` single-def fast path, matching `mergeWithAspectMeta` behavior

## Test plan

- [x] 771/771 CI tests pass
- [x] andrewix config evaluates correctly — `fileSystems."/"` and `"/boot"` populated on andrew-laptop/andrew-pc
- [x] WSL hosts correctly have empty fileSystems (no `_nixos` dir)
- [x] `import-tree.test-no-import-for-missing-dir` passes (pathExists guard)
- [x] `den._.import-tree._.host` resolves correctly (no `attribute '_' missing`)

### DIFF
--- a/modules/aspects/batteries/import-tree.nix
+++ b/modules/aspects/batteries/import-tree.nix
@@ -1,5 +1,6 @@
 {
   inputs,
+  lib,
   den,
   ...
 }:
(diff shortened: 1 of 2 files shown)


modules/aspects/batteries/import-tree.nix
```diff
@@ -59,23 +60,29 @@
...
-  };
+  den.batteries.import-tree.__functor =
+    _: root:
+    let
+      # Scan for _<class> directories under root and emit per-class imports.
+      # This avoids depending on the scope's `class` argument, which the
+      # fx-pipeline only provides once per scope (not once per class).
+      rootStr = toString root;
+      exists = builtins.pathExists rootStr;
+      entries = if exists then builtins.readDir rootStr else { };
```
drupol:
```suggestion
      entries = lib.optionalAttrs exists (builtins.readDir rootStr);
```

😆

sini: Fixed — `lib.optionalAttrs` is cleaner.

drupol (approved):

PR #527, opened 2026-05-13
fix: nested includes scoping and content wrapper callability

sini:
## Summary

Three fixes for users migrating from `provides`/`_` to direct nesting:

**1. Nested sub-aspects no longer auto-walk.** They must be explicitly included to emit class modules. This matches provides behavior and makes activation explicit. `den.aspects.root = { includes = [ den.aspects.root.a ]; a.nixos...; b.nixos...; }` now correctly includes only `a`.

**2. Multi-def merge fixed at the source.** `aspectContentType.merge` now produces per-key merged values instead of lossy shallow-merge (`foldl' //`). When multiple modules define `den.aspects.system.base.nixos`, all definitions contribute. This eliminates `mergeContentValuesPerKey` from the pipeline — every consumer sees properly merged keys regardless of code path.

**3. Single-function content wrappers are callable.** `den.aspects.X.Y = fn:` now gets `__functor` so `(den.aspects.wm.gnome-autologin "benjamin")` works without `provides`.

### Changes

- `nix/lib/aspects/types.nix` — per-key merge in `aspectContentType.merge`, `__functor` for single-function wrappers
- `nix/lib/aspects/fx/handlers/compile-static.nix` — remove auto-walk (`nestedToWalk`, `emitNestedAspect`), net -80 lines
- `nix/lib/aspects/fx/content-util.nix` — remove now-unused `mergeContentValuesPerKey`
- 6 test files — update expectations, add explicit includes

### Known limitation

`nested-aspects.test-direct-nesting-basic` fails — entity-linked nested keys (`igloo.tux` where `tux` is a declared user) previously relied on auto-walk for activation. Policy-driven activation for these is a separate design topic.

## Test plan

- [x] 784/785 CI tests pass (1 known: entity-linked activation)
- [x] Multi-def merge: `deadbugs.nested-aspect-merge`, `deadbugs.content-wrapper-functor`
- [x] No auto-walk: `deadbugs.nested-includes-scoping`
- [x] Explicit includes: `nested-aspects`, `structural-detection`
- [x] Formatted with `just fmt`


sini:
Fixing another regression I found during implementation.

drupol (approved):

osatomic:
Found an issue with deeply nested sub-aspects that are defined in separate files.

```nix
test-something = denTest (
  { den, igloo, ... }:
  {
    # fails
    imports = [
      { den.aspects.root.sub1.sub2.a.nixos.environment.variables.FROM_A = "yes"; }
    ];
    # passes
    # den.aspects.root.sub1.sub2.a.nixos.environment.variables.FROM_A = "yes";

    den.aspects.root.sub1.sub2.b.nixos.environment.variables.FROM_B = "yes";

      den.aspects.root.sub1.sub2.a
      den.aspects.root.sub1.sub2.b
    ];

    expr = {
      hasA = igloo.environment.variables ? FROM_A;
      hasB = igloo.environment.variables ? FROM_B;
    };
    expected = {
      hasA = true;
      hasB = true;
    };
  }
);
```

sini:
> Found an issue with deeply nested sub-aspects that are defined in separate files.

image

Roger roger, on it.

sini:
The above commit is a band-aid that fixes this edge case, but the right solution is going to be a refactor of the wrapping mechanism. At the expense of some scope creep, I'm going to try and tackle the refactor.

sini:
Okay, I've confirmed that the scope for this refactor is going to actually scope creep to nearly the size of the already designed flake-aspect 2.0 library extraction effort scheduled for next week/after we do this point release so I'm going to leave the band-aid for now. We know what to replace, and we now have these tests to make sure we don't reintroduce the regression. Going to merge this.

PR #529, opened 2026-05-13
ci: bump nix-effects to den branch

vic:
Explicitly using `denful/nix-effects/den` branch instead of the `vic` branch. As part of less-vic is more.


vic:
I will investigate it later tonight or tomorrow. Old `vic` branch had no changes, so current code is not affected.

sini (approved):
my own change looks good to me

PR #530, opened 2026-05-13
fix: policy.when with hasAspect no longer cycles, policy.for accepts effect descriptors

### LINKED ISSUE #525: Infinite recursion with `den.lib.policy.when` and `<entity>.hasAspect`
I was attempting to configure a user depending whether the host contains an aspect. I attempt to do this with the `<entity>.hasAspect` function, but end up with infinite recursion. There is a reproduction repo at https://github.com/jadestar1864/den-sibling-bug. Don't mind the weird settings, they were supposed to be meant as markers. Here is a snippet of the reproduction test:
```nix
test-something = denTest (
  {
    den,
    lib,
    igloo, # igloo = nixosConfigurations.igloo.config
    tuxHm, # tuxHm = igloo.home-manager.users.tux
    ...
  }: {
    # replace <system> if you are reporting a bug in MacOS
    den.hosts.x86_64-linux.igloo.users.tux = {};

    den.aspects.my-aspect.nixos.programs.git.enable = true;

    den.aspects.igloo.includes = [den.aspects.my-aspect];

    den.aspects.tux.includes = [
      (den.lib.policy.when ({host, ...}: host.hasAspect den.aspects.git) {
        nixos.programs.git.config.init.defaultBranch = "custom-main";
      })
    ];
    expr = igloo.programs.git.config.init.defaultBranch;
    expected = "custom-main";
  }
);
```

sini:
## Summary

- `policy.for` and `policy.when` now accept policy effect descriptors (e.g. `policy.include`) and inline aspect attrsets as arguments, not just raw functions or existing policies
- Fixes crash: `policy.for entity (policy.include aspects.foo)` → "attempt to call something which is not a function but a set"
- Fixes #525: `policy.when ({host,...}: host.hasAspect ref) { ... }` → infinite recursion
- Guards are walk-order independent: deferred on first failure, re-evaluated against final pathSet

## How it works

**`policy.for`/`policy.when` accept effect descriptors** — `toInnerPolicy` helper coerces effect descriptors and inline attrsets into callable policies. Four-branch dispatch: existing policy → passthrough, effect descriptor → constant function, inline attrset → include effect, raw function → direct use.

**`policy.when` with non-function values emits conditional aspects** — instead of registering a policy (which fires during dispatch, causing the `hasAspect` → `config.resolved` cycle), inline aspects and effect descriptors become conditional aspects with `meta.guard`. These route through `compile-conditional`, which evaluates guards against the in-flight pathSet — no dependency on `config.resolved`.

**Entity-shaped guard context** — `compile-conditional` enriches the guard context with entity stubs (`host = { hasAspect }`, `user = { hasAspect }`) derived from scope handler keys. This lets predicates written as `({ host, ... }: host.hasAspect ref)` work without touching entity config.

**Deferred guard evaluation** — when a guard fails (aspect not yet in pathSet), the conditional is deferred via `defer-conditional` and re-evaluated at `resolve-children` boundary when the pathSet is complete. Still-failing guards produce tombstones. This makes guards walk-order independent.

## Changes

- `nix/lib/policy-effects.nix` — `toInnerPolicy` helper, `policy.when` conditional path, throws on non-include effect descriptors
- `nix/lib/aspects/fx/handlers/compile-conditional.nix` — `mkGuardCtx`, `mkPipelineHasAspect`, `deferConditional`, `deferConditionalHandler`, `drainConditionalsHandler`
- `nix/lib/aspects/fx/handlers/resolve-children.nix` — drains conditionals after children resolve
- `nix/lib/aspects/fx/handlers/state-util.nix` — defensive `or` in all three scoped helpers
- `nix/lib/aspects/fx/pipeline.nix` — new state field, handler registrations
- `nix/lib/schema-util.nix` — extracted `schemaEntityKindsSet` (was duplicated in 4 files)
- 4 files updated to use shared `schemaEntityKindsSet`

## Test plan

- [x] `test-for-include-fires` — `policy.for entity (policy.include ...)` delivers content
- [x] `test-for-include-suppressed` — non-matching entity suppresses
- [x] `test-when-include-fires` — `policy.when pred (policy.include ...)` works
- [x] `test-when-inline-aspect` — `policy.when pred { nixos... }` works
- [x] `test-hasaspect-guard-fires` — `host.hasAspect` in guard fires correctly
- [x] `test-hasaspect-guard-suppresses` — guard suppresses when aspect absent
- [x] `test-hasaspect-from-user-scope` — works from user aspect includes
- [x] `test-deferred-guard-fires-after-walk` — walk-order independence via deferral
- [x] `test-hasaspect-foranyclass` — `forAnyClass` variant works
- [x] Full CI: 795/795 passing

### DIFF
--- a/templates/ci/modules/features/policy-for-include.nix
+++ b/templates/ci/modules/features/policy-for-include.nix
@@ -75,5 +75,25 @@
         expected = "from-zed";
       }
     );
+
+    # policy.when wrapping an inline aspect attrset should work.
+    test-when-inline-aspect = denTest (
+      { den, igloo, ... }:
+      let
+        inherit (den.lib) policy;
+      in
+      {
+        den.hosts.x86_64-linux.igloo.users.tux = { };
+
+        den.schema.host.includes = [
+          (policy.when (_: true) {
+            nixos.networking.hostName = "from-inline";
+          })
+        ];
+
+        expr = igloo.networking.hostName;
+        expected = "from-inline";
+      }
+    );
   };
 }
--- a/templates/ci/modules/features/policy-for-include.nix
+++ b/templates/ci/modules/features/policy-for-include.nix
@@ -1,0 +1,79 @@
+{ denTest, ... }:
+{
+  flake.tests.policy-for-include = {
+    # policy.for wrapping policy.include should work (was: "attempt to call
+    # something which is not a function but a set").
+    test-for-include-fires = denTest (
+      { den, config, ... }:
+      let
+        inherit (den) aspects;
+        inherit (den.lib) policy;
+        yalova = den.hosts.x86_64-linux.yalova;
+        yalovaConfig = config.flake.nixosConfigurations.yalova.config;
+      in
+      {
+        den.hosts.x86_64-linux.igloo.users.tux = { };
+        den.hosts.x86_64-linux.yalova.users.tux = { };
+
+        den.schema.host.includes = [
+          (policy.for yalova (policy.include aspects.zed-editor))
+        ];
+
+        den.aspects.zed-editor = {
+          nixos.networking.hostName = "from-zed";
+        };
+
+        expr = yalovaConfig.networking.hostName;
+        expected = "from-zed";
+      }
+    );
+
+    # policy.for on a non-matching entity should suppress the include.
+    test-for-include-suppressed = denTest (
+      { den, igloo, ... }:
+      let
+        inherit (den) aspects;
+        inherit (den.lib) policy;
+        yalova = den.hosts.x86_64-linux.yalova;
+      in
+      {
+        den.hosts.x86_64-linux.igloo.users.tux = { };
+        den.hosts.x86_64-linux.yalova.users.tux = { };
+
+        den.schema.host.includes = [
+          (policy.for yalova (policy.include aspects.zed-editor))
+        ];
+
+        den.aspects.zed-editor = {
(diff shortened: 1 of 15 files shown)


templates/ci/modules/features/policy-for-include.nix
```diff
@@ -0,0 +1,79 @@
...
+    test-for-include-fires = denTest (
+      { den, config, ... }:
+      let
+        inherit (den) aspects;
+        inherit (den.lib) policy;
+        yalova = den.hosts.x86_64-linux.yalova;
+        yalovaConfig = config.flake.nixosConfigurations.yalova.config;
+      in
+      {
+        den.hosts.x86_64-linux.igloo.users.tux = { };
```
theutz, line 15: What's igloo doing here?

templates/ci/modules/features/policy-for-include.nix
```diff
@@ -0,0 +1,79 @@
...
+        den.aspects.zed-editor = {
+          nixos.networking.hostName = "from-zed";
+        };
+
+        expr = yalovaConfig.networking.hostName;
+        expected = "from-zed";
+      }
+    );
+
+    # policy.for on a non-matching entity should suppress the include.
```
theutz, line 31: Ah.

theutz (approved):
LGTM! Thanks!

theutz (approved):

PR #531, opened 2026-05-13
feat: auto-label allow-ci for collaborators and add CODEOWNERS

sini:
## Summary
- Adds `auto-allow-ci` workflow using `pull_request_target` that checks if the PR author has write/admin collaborator access and automatically adds the `allow-ci` label
- Adds `CODEOWNERS` to auto-request reviews from @vic and @sini for changes to `nix/` and `modules/`

## Test plan
- [ ] Open a test PR from a collaborator account — verify `allow-ci` label is added automatically
- [ ] Open a test PR from a non-collaborator — verify no label is added
- [ ] Modify a file in `nix/` or `modules/` — verify review is requested from vic and sini

PR #532, opened 2026-05-13
fix: auto-allow-ci permissions and trigger

sini:
## Summary
- Adds `pull-requests: write` permission so `GITHUB_TOKEN` can add labels
- Removes `synchronize` trigger so the workflow only runs on PR open/reopen, not on every push

Fixes the `Resource not accessible by integration` error from #531.


sini:
Merging directly since again -- the only way to test is live. >.<

sini:
@vic this one worked by the way :exclamation:

PR #533, opened 2026-05-13
fix: user-scope detection for home-env batteries and diagram null guard

sini:
## Summary

- **User-scope detection for home-env batteries**: The host-scope battery policy relied on `host.users` to detect class membership, which fails when users come from external registries resolved via policies (e.g. fleet-demo). Added `userDetect` — a user-scope policy on `den.schema.user.includes` that fires at `{ host, user }` context regardless of how the user was resolved. Applied to all three batteries: homeManager, hjem, maid.

- **Guard hjem/maid against missing inputs**: `hjem.nix` and `maid.nix` unconditionally accessed `inputs.hjem`/`inputs.nix-maid` in `getModule`, crashing any flake without those inputs. Now throws an actionable error message instead.

- **Fix diagram null coercion**: `entityInstance` is explicitly set to `null` in graph entries, but `text.nix` used `or ""` which only catches missing attrs. Changed to proper null-aware checks matching the existing `instOf` pattern from `graph.nix`.

## Test plan

- [x] 786/786 CI tests pass
- [x] `nix eval --override-input den . ./templates/fleet-demo#den` evaluates cleanly
- [x] `nix flake check --override-input den . ./templates/minimal` passes

### DIFF
--- a/nix/lib/diag/text.nix
+++ b/nix/lib/diag/text.nix
@@ -7,6 +7,10 @@
 # All functions return plain strings (not derivations).
 { lib }:
 let
+  # entityInstance is explicitly null in some graph constructors;
+  # Nix `or` only catches missing attrs, not null values.
+  instOf = default: e: if e.entityInstance or null == null then default else e.entityInstance;
+
   # --- Table helpers ---
 
   # Render a markdown table from headers and rows.
@@ -182,7 +186,7 @@ let
       aspectsByHost = lib.foldl' (
         acc: e:
         let
-          inst = e.entityInstance or "";
+          inst = instOf "" e;
           parts = lib.splitString ":" inst;
           kind = builtins.head parts;
           name = if builtins.length parts > 1 then lib.concatStringsSep ":" (lib.tail parts) else "";
@@ -278,7 +282,7 @@ let
       instanceGroups = lib.foldl' (
         acc: n:
         let
-          inst = n.entityInstance or "unscoped";
+          inst = instOf "unscoped" n;
         in
         acc // { ${inst} = (acc.${inst} or [ ]) ++ [ n ]; }
       ) { } userAspects;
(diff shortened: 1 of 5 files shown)


theutz:
Fixed my issue!

nix/lib/diag/text.nix
```diff
@@ -182,7 +182,7 @@ let
       aspectsByHost = lib.foldl' (
         acc: e:
         let
-          inst = e.entityInstance or "";
+          inst = if e.entityInstance or null == null then "" else e.entityInstance;
```
theutz: `inst = e.entityInstance or ""` ?

nix/lib/diag/text.nix
```diff
@@ -278,7 +278,7 @@ let
       instanceGroups = lib.foldl' (
         acc: n:
         let
-          inst = n.entityInstance or "unscoped";
+          inst = if n.entityInstance or null == null then "unscoped" else n.entityInstance;
```
theutz: `inst = n.entityInstance or "unscoped";` ?

nix/lib/diag/text.nix
```diff
@@ -299,7 +299,7 @@ let
         (
           if n.isParametric or false then "yes (${lib.concatStringsSep ", " (n.fnArgNames or [ ])})" else "no"
         )
-        (n.entityInstance or "—")
+        (if n.entityInstance or null == null then "—" else n.entityInstance)
```
theutz: n.entityInstance or "-"

nix/lib/diag/text.nix
sini: `or` only triggers for missing attrs, not explicit `null`. `entityInstance` is set to `null` in `graph.nix:35,71`, so `e.entityInstance or ""` evaluates to `null` — which then crashes `lib.splitString`. This matches the existing `instOf` helper pattern at `graph.nix:148`.

nix/lib/diag/text.nix
sini: Same as above — `entityInstance` is explicitly `null` in several graph constructors, so `or` does not catch it.

nix/lib/diag/text.nix
sini: Same — explicit `null` passes through `or` unchanged.

nix/lib/diag/text.nix
theutz: AH! The more you know... 😄

theutz (approved):

PR #536, opened 2026-05-14
feat: ._ shorthand to include all child aspects

sini:
## Summary

- `den.aspects.foo._` now acts as a synthetic aspect that auto-includes all immediate child aspect keys of `foo`
- Enables `includes = [ den.aspects.foo._ ]` instead of enumerating each child explicitly
- Synthetic keys hidden behind `__functor` so existing `lib.attrValues` patterns on `_` are unaffected
- Adds `excludes` to `structuralKeysSet` (pre-existing gap now surfaced)

## Implementation

**`types.nix`**: Both `mergeWithAspectMeta` and `mergeFunctions` compute child aspect keys (filtering out structural, class, pipe, forwarded, and `__`-prefixed keys) and build a synthetic aspect delivered via `__functor` on provides/_.

**`normalize.nix`**: `wrapFunctorChild` detects zero-arg functors returning aspect-shaped values and resolves them immediately.

**`key-classification.nix`**: Added `excludes` to `structuralKeysSet`.

## Test plan

- [x] 7 new tests in `include-children` suite (basic, empty, mixed, provides-coexistence, parametric, attrValues compat, no-grandchildren)
- [x] 801/801 full CI passes

### DIFF
--- a/nix/lib/aspects/fx/aspect/normalize.nix
+++ b/nix/lib/aspects/fx/aspect/normalize.nix
@@ -26,6 +26,21 @@ let
     in
     if builtins.isFunction innerFn && isSubmoduleFn innerFn then
       normalizeModuleFn innerFn
+    # Synthetic provides: zero-arg functor returning an aspect-shaped value.
+    # Resolve immediately so the pipeline sees name/includes directly.
+    else if builtins.isFunction innerFn && innerArgs == { } && !(child ? __args) then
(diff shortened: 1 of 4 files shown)


vic (approved):
TMS - tests make sense :+1:

theutz:
This is an EXCELLENT idea!

PR #541, opened 2026-05-14
fix: deferred conditional guards drain at correct scope level (#540)

### LINKED ISSUE #540: BUG: <entity>.hasAspect doesn't work when it's included at the first place
Related to https://github.com/denful/den/issues/525 https://github.com/denful/den/pull/530

> Guards are walk-order independent

Hi. Sorry,  i just found a bug related to the PR. It's not work when the conditional is in the first place. It's a minor problem, but I'll assume this is a bug because it doesn't match what's described in the PR.

Here is the reproduction code (ref: https://github.com/fmway/den-policy-conditional-bogus)

```nix

  test-conditional-in-the-first-place = denTest (
      { den, igloo, ... }:
      {

          den.aspects.conditional
          den.aspects.base
        ];

        den.aspects.base.includes = [
          den.aspects.git
        ];

        den.aspects.conditional = {
          includes = [
            (den.lib.policy.when ({ host, ... }: host.hasAspect den.aspects.git) {
              nixos.environment.variables.GIT_ENABLED = "true";
            })
          ];
        };
        den.aspects.git.nixos.programs.git.enable = true;

        expr = igloo.environment.variables ? GIT_ENABLED;
        expected = true;
      }
    );
```
even included in `den.schema` it still fails when the the conditional and the entity (or parent) are at the same level.

e.g:
 ```nix
den.schema.host.includes = [
  den.aspects.conditional 
  den.aspects.base
];
```

sini:
## Summary

- Fixes #540: `policy.when` with `hasAspect` failed when the conditional was the first include and its dependency was a later sibling
- `drain-conditionals` was firing inside every `resolve-children` call, including nested aspects — the guard re-evaluated before siblings were walked
- Now drains only at entity-level aspects (`__entityKind`) or root scope aspects (`currentScope == rootScopeId`)

## Root cause

The conditional's `resolve-children` drained deferred guards before the parent finished walking all siblings. The pathSet was incomplete, so the guard failed and tombstoned permanently.

## Test plan

- [x] `test-conditional-before-dependency` — conditional first, dependency second in aspect includes
- [x] `test-conditional-same-level-schema` — same pattern via schema includes
- [x] Full CI: 805/805 passing (including all prior #525/#530 tests)

Co-Authored-by: @fmway who provided amazing tests

### DIFF
--- a/nix/lib/aspects/fx/handlers/compile-conditional.nix
+++ b/nix/lib/aspects/fx/handlers/compile-conditional.nix
@@ -1,7 +1,8 @@
 # Effect handler: compile-conditional
-# Evaluates guard against path-set, emits includes or tombstones.
-# Guards that fail are deferred for re-evaluation when the pathSet grows;
-# drain-conditionals resolves them at entity boundary.
+# Evaluates guards with exclude-aware hasAspect. Guards that fail are
+# deferred for re-evaluation when the pathSet grows (drain-conditionals).
+# Constraint registry is read eagerly — emitPolicyEffectsThen registers
+# excludes before processing includes, so guards see per-scope excludes.
 {
   lib,
   den,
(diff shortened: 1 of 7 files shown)


vic (approved):

fmway:
Great. Thanks for breakfast 🍞!

Issue #500, opened 2026-05-09
Recent Migration Related Issues/Feedback Mega-thread


sini:
We recently landed a very massive architecture shift with a number of API changes.

There are new docs at https://den.denful.dev/

There's a migration guide for those seeing `ctx` related warnings or errors: https://den.denful.dev/guides/migrate-ctx/

You no longer need to use den.lib.perHost/perUser/etc -- aspects are functions and can handle bare resolution, just pass `{host}` `{user}` or `{host,user}`. These fields are now also exposed directly to classes, and you only need them on the aspect if you wanted to provide different class modules by user.

We have a new rich policy system, it's in the documentation and an example is provided in the fleet-demo and terranix-demo templates of some usecases. Richer examples will come later.

Please share any issues here. I'm on-call `9AM-10PM PST` or `4PM - 5AM UTC` to offer support. I'll be following issues here on Github, but the fastest way to get my attention and get personalized direct debugging help is on our [Matrix channel](https://matrix.to/#/#den-lib:matrix.org) where I'm spending honestly too much of my day.

My goal is to make sure any issues you encounter are resolved ASAP.

gabrielsoldani:
The [old docs](https://den.oeiuwq.com/) now 404s, could you maybe set it up so it redirects to the new docs?

Search engines now bring up a ton of dead links when you search for den, and they haven't indexed the new docs yet, so I had to rely on GitHub's redirect from the old repo URL to this one to find out where you all went 😅

sini:
@vic anything you can do with regard to the pre- denful.dev routing?

vic:
I will forward *.oeiuwq.com domains into *.denful.dev

blemouzy:
Hello,

> We recently landed a very massive architecture shift with a number of API changes.

Thanks for all this amazing work!

> Please share any issues here.

Migrating `ctx` works fine but I have some issue during `den.aspects.X._.Y` to `den.aspects.X.Y` migration.

For example:

```nix
{
  den.aspects.system._.base.nixos =
    { pkgs, ... }:
    {
      environment.systemPackages = with pkgs; [
        fastfetch
      ];
    };
  
  den.aspects.system._.base.nixos =
    { pkgs, ... }:
    {
      fonts.packages = with pkgs; [
        nerd-fonts.dejavu-sans-mono
      ];
    };
}
```

works but

```nix
{
  den.aspects.system.base.nixos =
    { pkgs, ... }:
    {
      environment.systemPackages = with pkgs; [
[...]

PR #542, opened 2026-05-14
fix: excludes work on nested keys without _ prefix

sini:
## Summary

- Nested keys defined without `_` prefix (e.g., `my-aspect.sub-aspect1`) produce content wrappers `{ __contentValues, __provider }` with no `name` field
- When passed to `excludes`, `identity.key` produced `"<anon>"` instead of the actual key name, so the constraint never matched the aspect during the tree walk
- Fix: normalize content wrappers in `registerConstraints` by extracting `name` from `__provider` — the same normalization `wrapChild` does in `normalize.nix`

## Root cause

The aspect type system produces different shapes for nested keys vs `_`-prefixed sub-aspects:
- `den.aspects.foo.bar` → `{ __contentValues, __provider }` (content wrapper, no `name`)
- `den.aspects.foo._.bar` → `{ name = "bar"; meta.provider = [...]; ... }` (full aspect)

The constraint system computed identity on the raw wrapper before normalization injected the name.

## Test plan

- [x] `test-nested-key-exclude` — excludes both `_`-prefixed and non-prefixed sub-aspects
- [x] Full CI: 809/809

### DIFF
--- a/nix/lib/aspects/fx/aspect/children.nix
+++ b/nix/lib/aspects/fx/aspect/children.nix
@@ -131,10 +131,26 @@ let
           [ rawHandleWith ]
         else
           [ ];
+      # Compute exclude identity, normalizing content wrappers that have
+      # __provider but no name (nested keys without _ prefix).
+      excludeIdentity =
+        ref:
+        if builtins.isAttrs ref && ref.__isPolicy or false then
+          ref.name
+        else if builtins.isAttrs ref && ref ? __provider && !(ref ? name) then
+          let
+            prov = ref.__provider;
+          in
+          identity.key {
+            name = if prov != [ ] then lib.last prov else "<anon>";
+            meta.provider = if prov != [ ] then lib.init prov else [ ];
+          }
+        else
+          identity.key ref;
       excludeList = map (ref: {
         type = "exclude";
         scope = "subtree";
-        identity = if builtins.isAttrs ref && ref.__isPolicy or false then ref.name else identity.key ref;
+        identity = excludeIdentity ref;
       }) rawExcludes;
       allConstraints = handleWithList ++ excludeList;
       owner = aspect.name or "<anon>";
--- a/templates/ci/modules/features/deadbugs/issue-nested-key-exclude.nix
+++ b/templates/ci/modules/features/deadbugs/issue-nested-key-exclude.nix
@@ -1,0 +1,39 @@
+{ denTest, ... }:
+{
+  flake.tests.issue-nested-key-exclude = {
+    # Excluding a nested key (sub-aspect defined without _ prefix) should
+    # work the same as excluding a provides sub-aspect.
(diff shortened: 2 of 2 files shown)


theutz:
Is this something that got surfaced by my dalliance with excludes?

sini:
> Is this something that got surfaced by my dalliance with excludes?

It was raised in #500

fmway:
Interesting, so basically den has 2 aspect, sub-aspect and full aspect. While sub-aspect is maybe an unknown class, but they will be act as an aspect when they're inside `includes`. And full aspect is truly an aspect, that can be define in a root, or inside `provides`.

 Am i right?

sini:
> Interesting, so basically den has 2 aspect, sub-aspect and full aspect. While sub-aspect is maybe an unknown class, but they will be act as an aspect when they're inside `includes`. And full aspect is truly an aspect, that can be define in a root, or inside `provides`.
> 
> Am i right?

Sort of, yes. The types will be changing in a couple weeks, but that's the current type system. Ideally the typesystem would just let us say aspect types have keys of (aspect | class/quirk | fn-to-aspect) but for a multitude of reasons (backwards compat + identity + we can only detect functions, not if it's a class or fn-to-aspect)  we're doing a bit of gymnastics.

We'll be replacing this with denful/flake-aspects 2.0 after we've got everything hammered out and hopefully be able to simplify it then

sini:
> Interesting, so basically den has 2 aspect, sub-aspect and full aspect. While sub-aspect is maybe an unknown class, but they will be act as an aspect when they're inside `includes`. And full aspect is truly an aspect, that can be define in a root, or inside `provides`.
> 
> Am i right?

@fmway since you seem genuinely interested, and since I don't think you're in our matrix chat, I just published this doc that's helpful for understanding pipeline resolution flow: https://gist.github.com/sini/e326de73fca4704d663336eae86e546c

This is subject to change in our next version, but this is what we just shipped.

fmway:
> @fmway since you seem genuinely interested, and since I don't think you're in our matrix chat, I just published this doc that's helpful for understanding pipeline resolution flow: https://gist.github.com/sini/e326de73fca4704d663336eae86e546c
> 
> This is subject to change in our next version, but this is what we just shipped.

Okay, thanks. I'll land on the matrix later. Sorry, i just asked in the wrong place.

sini:
> > @fmway since you seem genuinely interested, and since I don't think you're in our matrix chat, I just published this doc that's helpful for understanding pipeline resolution flow: https://gist.github.com/sini/e326de73fca4704d663336eae86e546c
> > This is subject to change in our next version, but this is what we just shipped.
> 
> Okay, thanks. I'll land on the matrix later. Sorry, i just asked in the wrong place.

You're fine, just happy you're engaging. It's the open-source process. :)

theutz (approved):

PR #544, opened 2026-05-15
feat: scoped-import-tree example

vic:
cc @theutz. Tested by doing:

```nix
cd templates/scoped-import-tree
nix repl
> :lf . 
> did-scoped.hasBrackets
```


theutz (approved):
Brilliant!

PR #559, opened 2026-05-19
fix: hm-host crash with standalone homes + homeManager forwarding

sini:
## Summary

- **policy.when arg safety**: wrappers now pre-check the predicate's required args against the dispatch context before calling it. Prevents crash when `{ host, ... }` predicates dispatch in non-host contexts (e.g., standalone home resolution via late dispatch).
- **Fold hmHostBridge into makeHomeEnv**: added `schemaIncludes` parameter so the battery's `policyFn` includes schema content at both host scope (nixos/darwin keys) and user sub-scopes (homeManager keys route via `userForward`). Eliminates the separate bridge, `hasHmUsers`, and conditional host includes entry.

Fixes three user-reported issues:
1. `schema.hm-host` crashes when a standalone `den.homes` exists alongside hosts with HM users
2. `schema.hm-host` includes with `homeManager` class keys are silently dropped (nixos keys worked, homeManager didn't)
3. `policy.when` with strict predicates crashes in cross-scope dispatch contexts

## Test plan

- [ ] 6 new regression tests in `deadbugs/` (policy-when arg safety, standalone crash, homeManager forwarding)
- [ ] 829/829 existing CI tests pass
- [ ] `nix develop -c just ci` clean

### DIFF
--- a/modules/aspects/batteries/home-manager.nix
+++ b/modules/aspects/batteries/home-manager.nix
@@ -6,9 +6,7 @@
   ...
 }:
 let
-  inherit (den.lib.home-env) makeHomeEnv mkDetectHost;
-
-  result = makeHomeEnv {
+  result = den.lib.home-env.makeHomeEnv {
     className = "homeManager";
     ctxName = "hm";
     optionPath = "home-manager";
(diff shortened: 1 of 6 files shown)


drupol (commented):
Tests are passing, let's merge it.

theutz (approved):

PR #560, opened 2026-05-19
fix: inline permission check in allow-ci job

sini:
## Summary

- **Root cause**: `auto-allow-ci` adds the `allow-ci` label via `GITHUB_TOKEN`, but actions using `GITHUB_TOKEN` [do not trigger new workflow runs](https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow). The `labeled` event never fires for `test.yml`, so `allow-ci` is always skipped on the first (and only) run.
- **Fix**: Inline the collaborator permission check directly into the `allow-ci` job. When the label isn't present, the job checks `collaborators/{user}/permission` via the API. The label is kept as a fast-path and manual override for external contributor PRs.

## Behavior matrix

| Context | Result |
|---------|--------|
| Push to main | Step skipped, job succeeds |
| PR with `allow-ci` label | Step skipped, job succeeds |
| PR without label, author has write/admin | Step runs, permission passes, job succeeds |
| PR without label, external contributor | Step runs, permission fails, downstream blocked |

## Test plan

- [ ] Open a collaborator PR without manually adding label → template checks should run
- [ ] Verify `allow-ci` label still works as manual override

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -47,9 +47,20 @@ jobs:
     needs: [non-draft]
     name: allow-ci
     runs-on: ubuntu-latest
-    if: ${{github.ref == 'refs/heads/main' || contains( github.event.pull_request.labels.*.name, 'allow-ci')}}
+    # Label check kept as fast-path; permission check runs when label is absent.
+    # auto-allow-ci adds the label via GITHUB_TOKEN which does not re-trigger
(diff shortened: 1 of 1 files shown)


sini:
Can't test this not on main... so just going to merge it.

PR #563, opened 2026-05-21
feat: gen-schema entity port + resolver decoupling + delivery-edge unification

sini:
## What this delivers

This branch began as the gen-schema entity port and grew into a five-part stabilization of den's resolver and delivery pipeline. Each part builds on the last and is covered by CI. **1023/1023 tests** (up from 825 at branch start).

### 1. gen-schema entity port (the original scope)
- `den.schema` now uses gen-schema's `mkSchemaOption`/`mkSchemaEntryType` (sidecar extraction, computed fields, `__functor` wrapping); `resolvedCtxModule` extracted to shared `_types.nix`.
- Entities own their `id_hash` via gen-schema `mkInstanceType` (context-free kind+name identity).
- Flat declaration forms alongside the legacy two-level form for both hosts and homes (`den.hosts.igloo = { … }`, `den.homes."tux@igloo" = { … }`), preprocessed to the canonical shape — all consumers see the unchanged `{ system.name = entity }`.
- `den.reservedKeys` (user-extensible structural keys); `settings` reserved as structural.

### 2. Isolation + delivered-child host (microvm guests)
- Kind-level `isolated` marker, recorded per-scope at scope creation.
- Isolation-aware subtree collection (an isolated entity is its own delivery root).
- Routes deliver re-instantiating payloads **verbatim and keyed** (microvm `*.config` re-runs eval with the base module list); `reinstantiate` kept as a core route flag.
- Projected (in-context) `hasAspect` re-keyed by entity `id_hash` rather than scope-string, fixing a false-negative (`core.impermanence` → identity paths) for ancestor-nested hosts.
- *(Note: the intermediate `guest-os` class — added and removed within this branch, never in a release — left no surface; isolated guests carry honest `nixos` identity.)*

### 3. Resolver decoupling — binding half
Purifies *who binds what, where emission lands*, removing the host/user-specific machinery the resolver core carried (zero `host`/`user`/`homeManager` literals remain in the fx resolver core).
- A pure arg classifier over the schema entity DAG; **synchronous relationship fan-out** for descendant entity args (replaces the cross-scope deferral carrier).
- Kind-generic spawn materialization and a generic parent-chain root-owner lookup.
- Fixes the host-aspect→all-users leak (#609): a host-scope `{ user, … }` homeManager aspect no longer leaks to every user.
- `den.lib.perHost`/`perUser`/`perHome` are **kept as deprecated shims** (they shipped in main) — now thin aliases over the new rule.

### 4. Delivery-edge unification — delivery half
Collapses nine ad-hoc delivery mechanisms into one `(S, T, P, M)` edge algebra: a single mechanism-free mode switch, one constructor per mechanism, an explicit per-root context projection, and a record-level toposort for ordering — production now materializes delivery as **one toposorted edge fold per root**, replacing the four hand-threaded phase-fold sites.
- New user-facing **`deliver { from; to; at ? []; mode ? "merge"; }`** primitive; `route` and `provides` retained as permanent thin shims over it (signatures unchanged).
- `findHostScopeId`'s name-infix heuristic dissolved into an exact scope-creation link.
- The four delivery sites (`fxResolveFull`, `fxResolveImports`, the per-host re-walk, the spawn re-entry) all fold the single `materializeUnified` engine over `topoSortEdges(unifiedEdges)`; the per-mechanism phase orchestration is deleted. The toposort is stable (construction-order tiebreak), so byte-equivalence to the prior phase order holds.
- The **edge-trace is the production edge set**: `edgeTrace` renders the *captured* edges the production fold dispatched (oracle ≡ production by construction, no drift surface), with `legacyEdgeTrace` retained as the differential's "before". A production-vs-legacy differential proves the prior spawn `rewalk` undercount + dedup-suppressed twins are eliminated.
- A frozen, versioned **edge-schema parity contract** + `assertEdgeParity` + a parity corpus, so den-hoag (v2) can be validated against the v1 production edge set as a pure diff. Schema/runbook live in the den-architecture papers.
- One intended runtime fix: cross-host config resolution over a pipe-consuming peer now builds over assembled contexts (was a latent `attribute missing` throw; witness test added).

### 5. Constraint-registry unification + sibling-exclusion fixes (#613)
The `#613` scope-isolation fix (conditional-guard `hasAspect`) is extended to the exclude/substitute **application** path, and the constraint registry is unified.
- `aspects.X.excludes` and policy-name excludes no longer **leak across sibling entities** (an eval-order-dependent bug: a host walked earlier suppressed a sibling that included the same aspect/policy). Both flavors are now entity-scoped (scope + ancestors).
- The fleet-wide `flatConstraintRegistry` is **deleted**: all three readers (`check-constraint`, the policy dispatch, and the late-policy dispatch) go through one scope+ancestor lookup. The late-policy dispatch scopes to the *sibling it emits for*, so schema-tier excludes (`den.schema.KIND.excludes`) still reach their descendants.
- Dedup: one cycle-guarded `foldScopeAncestors` underlies the constraint walk and the guard pathSet walk (the latter also gains the full cycle guard).

## Breaking change

**One** breaking change versus the merge base — a **semantic** change, no API removals:

- **Entity-arg binding rule (#609).** A parametric aspect destructuring an entity kind now follows: bind once at the emitting scope if in-context → else fan out class-locally over schema-DAG descendants → else **inert** (silently; incl. any entity-kind arg at the root/flake scope). Previously such args could be delivered cross-scope. Cross-entity delivery is now via `provides`/policy only. This applies uniformly to plain parametric functions **and** to the deprecated `perHost`/`perUser`/`perHome` shims (whose old self-suppression *was* the #609 bug). Documented in `explanation/parametric.mdx` (the rule + the silent-inert footgun) and `reference/lib-deprecated.mdx`.

No public API is removed: `perHost`/`perUser`/`perHome`, `route`, `provides`, the legacy two-level entity form — all retained (the first three as warning shims). Release notes will be drafted as `heads-up` GH discussions at release time per `maintainers.mdx`.

## Dependency note

den's root flake is a pure library (`outputs = _: import ./nix`). gen-schema is resolved via the CI template lock, pinned to `github:sini/gen-schema` — the same personal-repo + CI-lock-fallback pattern den already uses for nix-effects. Staying as-is for now.

## Verification

- den CI **1023/1023** (`nix develop -c just ci`).
- **Delivery-edge slice (part 4, deliver-API era) is byte-identical downstream**: a real nix-config fleet host's full system closure builds to the **same derivation hash** at the delivery-edge baseline vs that point — not flake-source-only, the same store path. Transitively covers agenix identity paths, the microvm guest, impermanence, and home-manager delivery.
- **The toposort production switch is in-repo byte-stable**: each of the four fold sites was verified against the `entity-isolation` suite (real `evalModules` output) + full CI with no diffs; the oracle-vs-production differential gates the edge-set faithfulness. A nix-config host system-closure-hash re-verification across the swap is the only out-of-repo step, taken at the den input bump.
- delivery-edge contract fixtures, `deliver` API suite, the edge-toposort gates (`fx-materialize-unified`, `fx-unified-edges`, `fx-edge-unification-gate`, `fx-oracle-production-differential`, `fx-edge-parity`), and the #613 sibling-exclusion regressions all green.

## Status

Ready for review. Deliverability audit complete (dependency resolution, breaking-change docs, leftover-scaffolding sweep, templates): gen-schema stays at `sini` (matches nix-effects); flake-parts-modules template pin fixed; `perHost` shim restored; docs corrected. Specs, plans, the per-task deviation ledger, and the frozen parity contract live in the den-architecture papers; a net-behavior-change summary is published as a gist.


fmway (dismissed):
Looks fine for me. But, you may need to edit the doc comments in a few files (`nix/lib/aspects/fx/edges/edge.nix`, `nix/lib/aspects/fx/edges/materialize.nix`, `nix/lib/fx/resolve.nix`, `nix/lib/fx/spawn-node`, and `templates/ci/modules/deadbugs/bprime-basedrain-crosshost.nix`). I just found a few comments containing "Task ..." that refer to Agent-related matters.

docs/src/content/docs/guides/home-manager.mdx
```diff
@@ -236,3 +236,16 @@ den.hosts.x86_64-linux.laptop = {
...
+Earlier releases let a `{ user, ... }` aspect at host scope leak its
+`homeManager` content to each of the host's users. A host-scope parametric
+aspect now fans out over the users but emits class-locally **on the host**,
+where `homeManager` is inert -- so that content never reaches the users'
+Home Manager evaluation. This matches the resolver's emission rule: the bound
+user is the arg source, not the output target.
+
+If you were relying on that leak to push home content from a host-scope aspect,
+route it through an explicit `to-users` policy, `provides`, or the host-aspects
+battery instead.
```
fmway, line 251: What's the mean of `to-users` policy? is it a policy with the name `to-users`? Last time (on the main branch) i tried creating some policies with the same name (to-users),  but got auto-dedup and only left one policy

sini: This is a feature of the latent mutual-providers support -- to-host and to-users project the aspects across scope, in addition to 'entity-name'.

sini:
Addressed the comment consistency. I'm going to do a dead-code cleanup pass and one more validation check on named providers since we changed from basic name matching to actually using the entity gen-schema id_hash.

nix/lib/entities/_types.nix
```diff
@@ -22,6 +23,16 @@ let
...
 
+  # Recursive merge without forcing leaf values. Unlike lib.types.anything this
+  # does not inspect values deeply (no mapAttrsRecursiveCond), avoiding infinite
+  # recursion when values reference other options (e.g. den.aspects).
+  deepMergeAttrs = lib.mkOptionType {
+    name = "deepMergeAttrs";
+    description = "recursively merged attribute set";
+    check = builtins.isAttrs;
+    merge = _loc: defs: builtins.foldl' (acc: def: lib.recursiveUpdate acc def.value) { } defs;
+  };
```
fmway, line 34:
```suggestion
  deepMergeAttrs = lib.types.attrsOf (lib.types.serializableValueWith { typeName = "anything"; });
```

Benefit: user can use overrides (mkOverride, mkAfter, mkBefore, mkDefault, ...)

fmway (dismissed):

nix/lib/entities/_types.nix
sini: I don't believe that's correct with the way these modules eval/get merged. I'll verify.

sini: Verified — the suggestion doesn't hold here. `serializableValueWith` builds `oneOf [bool int float str path (attrsOf valueType) (listOf valueType)]`, so its `check` forces every leaf to dispatch on type and rejects anything outside that set. Aspect declarations carry **function**-valued leaves (parametric `__args`, `__functor`), so the type throws on them. That's also exactly the deep-forcing the existing comment warns about — leaves can reference other options (`den.aspects`), and forcing them risks infinite recursion. `deepMergeAttrs` uses `check = isAttrs` + a non-forcing `recursiveUpdate` precisely to avoid both.

Minimal regression case:

```nix
let
  inherit (lib) types;
  deepMergeAttrs = lib.mkOptionType {
    name = "deepMergeAttrs";
    check = builtins.isAttrs;
    merge = _loc: defs: builtins.foldl' (acc: def: lib.recursiveUpdate acc def.value) { } defs;
  };
  current   = types.attrsOf (types.submodule { freeformType = deepMergeAttrs; });
  suggested = types.attrsOf (types.serializableValueWith { typeName = "anything"; });

  ev = t: v: (lib.evalModules {
    modules = [ { options.x = lib.mkOption { type = t; }; } { config.x = v; } ];
  }).config.x;

  # aspect-shaped: a function-valued leaf (where parametric __args / __functor live)
  aspectVal = { igloo = { nixos.networking.hostName = "h"; mk = arg: arg; }; };
in {
  current   = builtins.tryEval (builtins.deepSeq ((ev current   aspectVal).igloo.mk 1) "OK");
  suggested = builtins.tryEval (builtins.deepSeq ((ev suggested aspectVal).igloo.mk 1) "OK");
}
# => { current = { success = true; value = "OK"; };
#      suggested = { success = false; value = false; }; }   # throws on the function leaf
```

The override ergonomics you're after (`mkForce`/`mkMerge`/priorities) are a real gap — but a serializable type can't deliver them here since it can't admit functions and can't force leaves safely. If we want priority support on entity decls it'd need a custom merge that honors `mkOverride`/`mkMerge` while keeping `check = isAttrs` and not forcing leaf values. Happy to track that separately; keeping `deepMergeAttrs` as-is for this PR.

fmway: > ```nix
> suggested = types.attrsOf (types.serializableValueWith { typeName = "anything"; });
> ```
That shouldbe:

```nix
suggested = let
  deepMergeAttrs = types.attrsOf (types.serializableValueWith { typeName = "anything"; });
in types.attrsOf (types.submodule { freeformType = deepMergeAttrs; });

```

sini: Good catch on the apples-to-apples wrapping — I re-ran it with your exact construction (`serializableValueWith` as the submodule `freeformType`), and it still throws:

```nix
ff        = types.attrsOf (types.serializableValueWith { typeName = "anything"; });
suggested = types.attrsOf (types.submodule { freeformType = ff; });
# ...same ev/aspectVal as before...
# => { current = { success = true; ... }; suggested = { success = false; ... }; }
```

Concrete error:

```
error: A definition for option `x.igloo.mk' is not of type `anything value'.
Definition values:
- In `<unknown-file>': <function>
```

The wrapping doesn't change the outcome because the submodule `freeformType` only decides *which type handles undeclared keys* — the leaf values still flow through `serializableValueWith`'s `check`, and that check rejects a function at any depth (it's `oneOf [bool int float str path attrsOf listOf]`, no lambda branch). So a function-valued leaf like `mk` fails regardless of how many `attrsOf`/`submodule` layers sit above it.

`deepMergeAttrs`'s `check = isAttrs` only gates the *container*, never the leaves, which is why functions pass through untouched. That's the property we need here. Still happy to look at a priority-honoring custom merge separately, but the serializable route can't be the substitute.

fmway: Okay, nevermind then 😂. Sorry just wasted your time

sini: You're good. Keeping me honest; and you get to understand a bit about why this is the way that it is. :)

sini: This is part of the same mechanism that lets us have den.aspects.that.are.deeply.nested.

fmway (approved):

Issue #568, opened 2026-05-24
Flake parts migration documentation incomplete


tdresser:
The instructions in https://den.denful.dev/guides/from-flake-to-den/ don't seem to work with a flake parts setup.

In particular, calling `evalModules` with `inputs.den.flakeOutputs.flake` tries to write to `perSystem`, but `perSystem` isn't defined. 

I think den is doing some `inputs` snooping ([here](https://github.com/denful/den/blob/8f1a59448043677ac8bc7854348c1b8ee6889c0b/modules/outputs.nix#L9)), and thinks it's in a flake parts context, but then flake parts doesn't exist in the `evalModules` context.

I'm currently getting around this via passing 
```
 specialArgs = {
    inherit inputs;
    __denTest = true; 
 };
```
to the `evalModules` to prevent the flake parts sniffing, but it feels pretty hacky.

Maybe I should be trying to avoid passing full inputs to `evalModules`?

tdresser:
Oops, will bring this up in the appropriate forum.

PR #571, opened 2026-05-26
fix: remove __denTest hack

vic:
Fixes #570  (#568)


vic:
@sini looks like `__denTest` was no longer necessary by any template nor CI tests

sini (approved):

PR #573, opened 2026-05-29
fix: stop routing OS users into custom host classes

Scott-Guest:
## Bug
`den.batteries.os-user` is documented as forwarding the `user` class into OS-level `users.users.<name>` on NixOS/nix-Darwin hosts. However, the policy actually routes unconditionally into `host.class`, so custom host classes receive a `users.users.<name>` entry, even when no OS user module exists there.

## Changes
- Guard `den.policies.user-to-host` so it only emits routes for `host.class == "nixos"` or `"darwin"`.
- Add a regression test covering the custom host class case

## AI Assistance
Codex with GPT-5.5 xhigh was used to identify the issue, prepare this patch, draft the commit message, and draft this PR description. The work was guided, reviewed, and accepted by myself.

I, Scott Guest, explicitly acknowledge the following:

### DIFF
--- a/modules/aspects/batteries/os-user.nix
+++ b/modules/aspects/batteries/os-user.nix
@@ -33,7 +33,7 @@ in
 {
   den.classes.user.description = "Lightweight user environment forwarding to OS users.users";
 
-  # Built-in policy: route user class content to the host's OS at
+  # Built-in policy: route user class content to NixOS/nix-Darwin at
   # users.users.<userName>. Injects osConfig so user-class modules
   # can reference the parent NixOS/Darwin config.
   # The route's ensureEntry mechanism creates users.users.<name> even
@@ -41,17 +41,27 @@ in
   den.default.includes = [ den.policies.user-to-host ];
 
   den.policies.user-to-host =
-    { user, host, ... }:
-    [
-      (den.lib.policy.route {
-        fromClass = "user";
-        intoClass = host.class;
-        path = [
-          "users"
-          "users"
-          user.userName
-        ];
-        adaptArgs = args: args // { osConfig = args.config; };
-      })
-    ];
+    den.lib.policy.when
+      (
+        { user, host, ... }:
+        builtins.elem host.class [
+          "nixos"
+          "darwin"
+        ]
+      )
+      (
+        { user, host, ... }:
+        [
+          (den.lib.policy.route {
+            fromClass = "user";
+            intoClass = host.class;
+            path = [
+              "users"
+              "users"
+              user.userName
+            ];
+            adaptArgs = args: args // { osConfig = args.config; };
+          })
+        ]
+      );
 }
--- a/templates/ci/modules/features/os-user-class.nix
+++ b/templates/ci/modules/features/os-user-class.nix
@@ -14,6 +14,47 @@
       }
     );
 
+    test-does-not-forward-user-description-to-custom-host = denTest (
+      {
+        config,
+        den,
+        lib,
+        ...
+      }:
+      let
+        mockCustomHostModule =
+          { lib, ... }:
+          {
+            config._module.freeformType = lib.types.lazyAttrsOf lib.types.unspecified;
+          };
+
+        igloo-custom-host = config.flake.iglooCustomHostConfigurations."igloo-custom-host";
+      in
+      {
+        den.classes.custom-host.description = "Custom host configuration";
+
+        den.hosts.x86_64-linux.igloo-custom-host = {
+          class = "custom-host";
+          intoAttr = [
+            "iglooCustomHostConfigurations"
+            "igloo-custom-host"
+          ];
+          instantiate =
+            { modules, ... }:
+            (lib.evalModules {
+              modules = [ mockCustomHostModule ] ++ modules;
+            }).config;
+        };
+
+        den.hosts.x86_64-linux.igloo-custom-host.users.tux = { };
+
+        den.aspects.tux.user.description = "pinguino";
+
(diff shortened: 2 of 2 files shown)


modules/aspects/batteries/os-user.nix
```diff
@@ -42,16 +42,21 @@ in
...
-        path = [
-          "users"
-          "users"
-          user.userName
-        ];
-        adaptArgs = args: args // { osConfig = args.config; };
-      })
-    ];
+    lib.optional
+      (builtins.elem host.class [
```
drupol: I think you might perhaps replace this with `when` attribute in the policy definition. To be checked. 

https://den.denful.dev/explanation/policy-activation/#denlibpolicywhen-predicate-policy

sini:
Can you describe the custom host usage pattern more? You can also just add this policy to den.defaults.excludes list to disable it.

modules/aspects/batteries/os-user.nix
Scott-Guest: Agreed, and I verified that it's equivalent. Amended.

modules/aspects/batteries/os-user.nix
```diff
@@ -33,25 +33,35 @@ in
...
-          "users"
-          user.userName
-        ];
-        adaptArgs = args: args // { osConfig = args.config; };
-      })
-    ];
+    den.lib.policy.when
+      (
+        { user, host, ... }:
+        builtins.elem host.class [
```
sini, line 47: this might break with wsl or some microvm configs; I would like to understand the actual problem that made this an issue before reviewing the solution to it

sini (changes requested):

Scott-Guest:
> Can you describe the custom host usage pattern more? You can also just add this policy to den.defaults.excludes list to disable it.

"custom" is a bit of a misnomer for my exact use case. I have `host.class = "systemManager"`, and System Manager does actually support `users.users`. Disabling the policy is fine as a workaround.

But regardless, I think the issue pointed at by this PR is real:

- `os-user` has documentation ([here](https://github.com/denful/den/blob/9dfaf7443af65e742f97b306dc4df1de181def3f/modules/aspects/batteries/os-user.nix#L8-L9) and [here](https://github.com/denful/den/blob/30ebfdc1b6b3e02f668b2559c0af75cbd1cfe086/docs/src/content/docs/guides/custom-classes.mdx?plain=1#L55-L56)) implying it's for NixOS/nix-Darwin hosts.
- `user-to-host` routes into whatever `host.class` is, so it can emit `users.users.<name>` even if `host.class` is some custom class that does not support `users.users`.
- From what I can gather, other batteries guard config to the specific supported classes, avoiding this sort of issue.

---

Looking at this deeper, my actual usage is running into a related issue that's closer to a UX bug, which this PR doesn't address. Say I have:
```nix
{
  den.hosts.x86_64-linux.workstation = {
    class = "systemManager";
    users.tux.classes = [ ]
  };
}
```

This config does not provide any explicit `den.aspects.tux.user = { ... }`, and it opts out of the defaulted `classes = [ "user" ]`, so I would expect to only get the Den-level user entity. I would not expect an OS-level `users.users.tux` entry.

Somewhat surprisingly though, System Manager fails with assertions like:
```text
Failed assertions:
- Exactly one of users.users.tux.isSystemUser and users.users.tux.isNormalUser must be set.
- users.users.tux.group is unset. This used to default to nogroup, but this is unsafe.
```

It turns out that, as mentioned in [this comment](https://github.com/denful/den/blob/9dfaf7443af65e742f97b306dc4df1de181def3f/modules/aspects/batteries/os-user.nix#L39-L40), `user-to-host` can create an empty placeholder OS-level `users.users.tux = { };` even when there is no real user class content to forward. This is useful for integrated Home Manager on NixOS/nix-Darwin because HM reads `config.users.users.<name>.name` / `.home` from the parent OS config.

This is surprising: `user-to-host` can silently add an OS-level `users.users.<name>` option (and an empty/incomplete one at that), solely from the existence of a Den user entity, even outside the HM-integrated-with-NixOS/nix-Darwin case where it may be necessary.

modules/aspects/batteries/os-user.nix
drupol: I think you can add the when attribute inside the policy you want to introduce, not outside and before it.

drupol: Oops, ignore this comment. That said, it might be an idea for @sini ?

E.g., we could do something like:

```nix
den.policies.user-to-host =
        { user, host, ... }:
        [
          (den.lib.policy.route {
            fromClass = "user";
            intoClass = host.class;
            path = [
              "users"
              "users"
              user.userName
            ];
            adaptArgs = args: args // { osConfig = args.config; };
            when = { user, host, ... }:
                builtins.elem host.class [ 
                  "nixos"
                  "darwin"
                ];
          })
        ]
```

modules/aspects/batteries/os-user.nix
```diff
@@ -33,25 +33,35 @@ in
...
+            intoClass = host.class;
+            path = [
+              "users"
+              "users"
+              user.userName
+            ];
+            adaptArgs = args: args // { osConfig = args.config; };
+          })
+        ]
+      );
```
fmway, line 66:
```suggestion
    den.lib.policy.route {
      fromClass = "user";
      intoClass = host.class;
      path = [
        "users"
        "users"
        user.userName
      ];
      adaptArgs = args: args // { osConfig = args.config; };
      guard = { options, ... }: options ? users.users;
     };
```

I'm wondering why not just use a guard?

EDIT:

Nevermind. I just read your real problem.

PR #574, opened 2026-05-29
fix: route wrapping merge semantics and forward ordering

### LINKED ISSUE #567: BUG: Class forwarder breaks when imported from mutual provider
Reproduction: https://github.com/musjj/forward-bug

Proof of regression: https://github.com/musjj/forward-bug/actions/runs/26342181368

### Description

When a class forwarder is imported from a mutual provider aspect, it will fail to forward certain classes. This is a regression from how it behaved in `v0.16.0`.

```nix
{ den, lib, ... }:
let
  nixClass =
    { class, aspect-chain }:
    den._.forward {
      each = [
        "nixos"
        "homeManager"
      ];
      fromClass = _: "nix";
      intoClass = lib.id;
      intoPath = _: [ "nix" ];
      fromAspect = _: lib.head aspect-chain;
      adaptArgs = lib.id;
    };
in
{

  den.default.includes = [ den._.mutual-provider ];

  den.aspects.tux.provides.igloo = {
    includes = [ nixClass ];
    nix.settings.core = 64;
  };
}
```

Here, `nix.settings.core` will be forwarded to `nixos` but not `homeManager`.

Related:
- https://github.com/denful/den/discussions/566

sini:
Closes #572, closes #567

## Summary

Three route application fixes in the fx pipeline:

**#572 — Multiple aspects defining the same flake output class clobber each other.** `wrapRouteModules` evaluated each source module independently through `nestWithAdaptArgs`, producing separate config definitions. The flake submodule's freeform `unspecified` type shallow-merged them with `//`, keeping only the last.

**#567 — Class forwarder with `each` breaks when imported from a mutual provider.** Forward routes like `nix→homeManager` registered after the HM battery route `homeManager→nixos`. The consumer fired first and collected from `homeManager` before the producer had written to it.

**Audit — `nestPlain` had the same shallow merge, freeform type was too permissive, and duplicate instantiate specs were silently dropped.**

## Changes

- **`route/wrap.nix`**: Combine all source modules into a single `evalModules` call for the `nestWithAdaptArgs` path. Align `nestPlain` to also use `evalModules` instead of `recursiveUpdate`. Introduce `mergeableType` — a custom freeform type that deep-merges attrsets, concatenates lists, and errors on conflicting scalar/derivation values (matching NixOS module system semantics).
- **`route/apply.nix`**: Add `topoSortRoutes` — partitions complex forward routes so producers (whose `intoClass` feeds another forward's `fromClass`) fire before consumers.
- **`resolve.nix`**: Add `lib.warnIf` when multiple instantiate specs target the same output path on the same system.
- **Regression tests**: 4 new tests across 2 files.

## Test plan

- [x] 833/833 CI tests pass
- [x] `flake-parts-modules` template passes (previously failed with `raw` freeform type)
- [x] All other templates pass `just check-all` (same results as main)
- [x] Formatting clean

### DIFF
--- a/nix/lib/aspects/fx/resolve.nix
+++ b/nix/lib/aspects/fx/resolve.nix
@@ -410,8 +410,16 @@ let
                   e // { path = basePath ++ [ "${baseName}@${e.system}" ]; }
                 ) entries
               else
-                # Same entity via multiple policy paths: keep last.
-                [ (lib.last entries) ];
+                # Same entity via multiple policy paths: deduplicate.
+                # Warn when more than one distinct instantiate spec targets the
+                # same output path on the same system — this can silently shadow
+                # one entity's configuration with another's.
+                let
+                  entry = lib.last entries;
+                in
+                lib.warnIf (builtins.length entries > 1)
+                  "den: multiple instantiate specs target ${builtins.concatStringsSep "." entry.path} on ${entry.system or "unknown"}; keeping last"
+                  [ entry ];
         in
         lib.concatLists (lib.mapAttrsToList resolve grouped);
 
--- a/templates/ci/modules/features/packages-merge.nix
+++ b/templates/ci/modules/features/packages-merge.nix
@@ -80,9 +80,8 @@
       }
     );
 
-    # Overlapping key: last included aspect wins for same package name
-    # (evalModules freeform merge with unspecified type).
-    test-overlapping-key-last-wins = denTest (
+    # Overlapping key: two aspects defining the same package name errors.
+    test-overlapping-key-errors = denTest (
       {
         config,
         den,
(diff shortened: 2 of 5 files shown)


drupol (approved):

templates/ci/modules/features/packages-merge.nix
```diff
@@ -0,0 +1,120 @@
...
+
+        expr = builtins.sort (a: b: a < b) (builtins.attrNames config.flake.packages.x86_64-linux);
+        expected = [
+          "alpha"
+          "beta"
+        ];
+      }
+    );
+
+    # Overlapping key: last included aspect wins for same package name
```
theutz: Wouldn't it be nicer UX to throw an error or at least an eval warning in this case?

sini: good point, heading back to the saltmines

drupol: Good idea indeed.

theutz (approved):
Pending passing CI, LGTM!

drupol (approved):

nix/lib/aspects/fx/resolve.nix
```diff
@@ -410,8 +410,16 @@ let
                   e // { path = basePath ++ [ "${baseName}@${e.system}" ]; }
                 ) entries
               else
-                # Same entity via multiple policy paths: keep last.
-                [ (lib.last entries) ];
+                # Same entity via multiple policy paths: deduplicate.
```
theutz, line 413: Image

PR #589, opened 2026-06-02
fix: backport aspect-pipeline fixes from entity-gen-schema-port

sini:
Backports the standalone aspect-pipeline bug fixes that accumulated on `feat/entity-gen-schema-port` so they can land on `main` independently of the gen-schema entity port (which is still in flight) and the den-diagram extraction (tracked in its own PR).

Most commits carry a `(cherry picked from commit ...)` trailer back to the source branch. **One commit (`e711eebe`) is net-new** — authored directly on this branch with no cherry-pick origin (see **Net-new** below).

## Generalized node-spawn + resolve-at-emitting-node (`c57ac6b8`)

A general primitive — **`spawnNode`** — materializes a child resolution node from *any* parent scope, threaded with the parent pipeline's resolved scope-tree state (parent + siblings), so the node's own `assemblePipes` re-derives inherited/collected pipe values with full fleet visibility. Paired with **resolve-at-emitting-node**: a pipeline-parametric pipe emit (`{ <arg>, ... }: …`) resolves to concrete data at its emitting node on *every* crossing (local / collected / exposed), never as a function; config-dependent emits stay deferred (`__configThunk`).

The mechanism is general — den-hoag `spawn`: a node with one read-only inherited edge, applicable to *any* parent→child entity relationship (fleet→host, environment→host, host→user, host→home). **Home extraction is the first and driving consumer:** it replaces the three isolated home-extraction sub-pipelines (`host-aspects` `resolveImports`; `makeHomeEnv` / `hm-host` `resolveEntity`) with `spawnNode`, fixing the originating bug where a host-aspects-projected ssh app saw only the local host, never the fleet peers.

**This supersedes the `home-env` chainCtx workaround** (`eea3d6b1`, under Fixes): that commit threaded ambient resolution-chain context into the isolated extraction as a stopgap; this change removes it — `environment` and fleet-collected pipe data now arrive structurally via the threaded scope-tree state.

- `assemble-pipes`: resolve-at-emitting-node on the collected and exposed crossings; config-dependent emits stay deferred.
- `spawn-node.nix`: the `spawnNode` primitive (re-walk for one class, merge parent state, own `assemblePipes`, class isolation).
- `policy.spawn` effect + `register-spawn` handler + drain augmentation: a deferred node spawn resolved post-walk.
- `route/apply.nix`: forward source resolves via `spawnNode` (`from = parent scope`).
- Bundles the **scopeParent-walk pipe inheritance** (pure-consumer scopes inherit a pipe's assembled value from the nearest policy-bound ancestor).

## Fixes
- **hasAspect — nested refs via `__provider` chain** (`5a3bf908`): `hasAspect` resolves nested aspect refs through the provider chain.
- **hasAspect — nested freeform refs at any depth** (`aa11ce89`): supports nested freeform aspect references regardless of depth, with provenance kept distinct.
- **deep-merge nested namespace children across files** (`ca4569af`): colliding nested namespaces declared across multiple files now deep-merge instead of last-write-wins.
- **navigated nested aspects get their own identity** (`beb2b493`): nested includes dedup correctly across scopes instead of sharing identity.
- **home-env resolution-chain ctx** (`eea3d6b1`): threads the resolution-chain context into home-manager extraction so parametric host quirks survive home projection. **Superseded by the node-spawn unification above** (`c57ac6b8`), which removes the chainCtx workaround in favor of `spawnNode`'s threaded scope-tree state.

- **policy dispatch — runtime-include policies scoped to their subtree** (`439c3dc5`): the late-sibling dispatch re-fired a sibling's own-include policies at every other sibling (entity-kind filter only), so a user opting into a battery — or any `{host,user}` policy a user registered — leaked to every other user on the host. Eligibility is now ancestor-or-self: a user's runtime includes stay in its subtree; host-registered provides still fan to all users. Fixes host-aspects projecting onto non-opt-in users (pre-existing leak).

## Net-new
- **namespaces: `_` provides bundle on namespace roots** (`e711eebe`): a namespace root now exposes `_` like an aspect, so `[ foo._ ]` bundles the namespace's **root-level aspects** — the container-level analog of `den.aspects.foo._`. The bundle is a flat map over the container's *direct* keys (structural keys `stages`/`schema`/`classes`/`_module`/`_` excluded); it does **not** recurse to enumerate nested aspects. Each root-level aspect is included whole, so its nested children still arrive — via normal include resolution, not via `_` descending — and the aspect-leaf `_` (e.g. `foo.app._`) remains a distinct, separately-computed bundle. Previously `foo._` threw `attribute '_' missing` because a namespace root is a container, not an aspect. Reported in discussion #588. Not a backport (no upstream commit); landed here per maintainer call.
- **namespaces: don't round-trip the synthetic `_` on re-import** (`95b6277e`): follow-up fix found in code review — the computed `_` was serialized into exported `denful` and collided with the read-only option when a re-imported namespace forced `._`. Now stripped on export and import.

## Tests
- **issue-583 mock fix** (`a2cb8708`): stops the issue-583 forwarding-overwrite test mock from colliding with nixpkgs `programs.atuin.flags`.
- **issue-588 regression suite** (`e711eebe`): 5 tests covering the namespace-root `_` bundle, the `den.aspects.foo._` baseline, explicit-list equivalence, aspect-leaf `ns.app._`, and structural-key exclusion.
- **home-extraction suite** (`c57ac6b8`): 8 tests — all-peers (collected, host-aspects projection), resolved-users (exposed), config-thunk deferral, in-tree ≡ threaded equivalency, server-host membership, and a complex-forward source-fallback guard.

## Excluded
- gen-schema entity port (separate PR).
- den-diagram / den-gram extraction (separate PR).
- `chore: update flake.lock` commits (bump gen-schema/den-diagram inputs not present on `main`).

## Verification
- Full CI: **863/863** passing (was 853/853; +8 from the new `home-extraction` suite).
- Affected suites isolated: `home-extraction` 8/8, `host-aspects` 10/10, `host-aspects-chain-ctx` 1/1 (parametric host quirk survives projection via the threaded scope-tree state, with chainCtx removed), `pipe-scope` 16/16, `pipes` 9/9, `deadbugs` 23/23, `has-aspect` 35/35.
- `just fmt` clean.

### DIFF
--- a/templates/ci/modules/features/deadbugs/host-aspects-chain-ctx.nix
+++ b/templates/ci/modules/features/deadbugs/host-aspects-chain-ctx.nix
@@ -1,0 +1,97 @@
+# Regression: the host-aspects battery re-resolves the host aspect tree for a
+# user's classes (e.g. homeManager) in an isolated sub-pipeline.  It seeded that
+# re-resolution with only { host, user }, dropping the resolution-chain context
+# the host scope actually carries (e.g. a parent `environment`).  A parametric
+# host quirk emit `{ environment, host, ... }: ...` was then stranded as a raw
+# function at the {host,user} projection scope, crashing any homeManager
+# consumer that read the pipe ("expected a set but found a function").
+#
+# Fix: from-host threads the ambient resolution-chain entity bindings into the
+# re-resolution, so parametric host aspects bind the same args they would at the
+# host scope.
+{ denTest, lib, ... }:
+{
+  flake.tests.host-aspects-chain-ctx = {
+
+    test-parametric-host-quirk-survives-home-projection = denTest (
+      {
+        den,
+        tuxHm,
+        lib,
+        ...
+      }:
+      {
+        # environment as a parent entity of host (mirrors a fleet topology).
+        den.schema.environment.isEntity = true;
+        den.schema.host.parent = "environment";
+
+        den.hosts.x86_64-linux.igloo.users.tux = { };
+
+        den.quirks.host-addrs.description = "Host address entries";
+
+        # Walk flake -> environment -> host, injecting `environment` into the
+        # host scope context.  Replaces the default per-system host walking.
+        den.policies.to-env = _: [
+          (den.lib.policy.resolve.to "environment" {
+            environment = {
+              name = "prod";
+              domain = "example.com";
+            };
+          })
+        ];
+        den.policies.env-to-hosts =
+          { environment, ... }:
+          lib.concatMap (
+            system:
+            lib.concatMap (
+              hostName:
+              let
+                host = den.hosts.${system}.${hostName};
+              in
+              [
+                (den.lib.policy.resolve.to "host" { inherit host; })
+                (den.lib.policy.instantiate host)
+              ]
+            ) (builtins.attrNames (den.hosts.${system} or { }))
+          ) (builtins.attrNames (den.hosts or { }));
+
+        den.schema.flake.includes = [ den.policies.to-env ];
+        den.schema.environment.includes = [ den.policies.env-to-hosts ];
+        den.schema.flake-system.excludes = [
+          den.policies.system-to-os-outputs
+          den.policies.system-to-hm-outputs
+        ];
+
+        # Host aspect tree: a parametric host-addrs emit requiring `environment`
+        # (a host-only ctx key), consumed by a homeManager aspect projected onto
+        # the user via the host-aspects battery.
+        den.aspects.igloo.includes = [
+          den.aspects.net-hosts
+          den.aspects.ssh-home
+        ];
+        den.aspects.net-hosts.host-addrs =
+          { environment, host, ... }:
+          {
+            hostname = host.name;
+            domain = environment.domain;
+          };
+        den.aspects.ssh-home.homeManager =
+          {
+            host-addrs,
+            lib,
+            ...
+          }:
+          {
+            home.sessionVariables.SSH_HOSTS = lib.concatMapStringsSep "," (
+              e: "${e.hostname}.${e.domain}"
+            ) host-addrs;
+          };
+
+        den.aspects.tux.includes = [ den._.host-aspects ];
+
+        expr = tuxHm.home.sessionVariables.SSH_HOSTS;
+        expected = "igloo.example.com";
+      }
+    );
+  };
+}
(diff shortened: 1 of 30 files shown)


templates/ci/modules/features/has-aspect.nix
```diff
@@ -619,6 +619,73 @@
...
+        };
+        expected = {
+          a-sub = true;
+          b-sub = false;
+        };
+      }
+    );
+
+    # Deeply nested freeform ref (3 levels).
+    test-H2-deeply-nested-freeform = denTest (
```
theutz, line 676: Is there any reason to believe that this might break at even deeper levels? If so, perhaps it would be good to make this deeper...

templates/ci/modules/features/deadbugs/nested-aspect-include-identity.nix
```diff
@@ -0,0 +1,52 @@
...
+          ))
+        ];
+
+        den.aspects.deep.grp.svc.nixos.boot.kernelParams = [ "den-dedup-marker" ];
+
+        # Host-scope inclusion (via a role the host includes).
+        den.aspects.gaming.includes = [ den.aspects.deep.grp.svc ];
+
+        # User-scope inclusion (entity-named aspect, auto-included by the policy).
+        den.aspects.igloo.tux.includes = [ den.aspects.deep.grp.svc ];
```
theutz, line 40: Can you really just add includes like that to a user? I didn't know `tux` on `igloo` would automatically forward to the user like that.

templates/ci/modules/features/deadbugs/host-aspects-chain-ctx.nix
theutz, line 1: I'll admit I don't fully understand this one.

theutz (approved):

templates/ci/modules/features/has-aspect.nix
sini: no, i'm using it with 5 levels deep in my config

templates/ci/modules/features/deadbugs/nested-aspect-include-identity.nix
sini: :smiling_imp:

templates/ci/modules/features/deadbugs/host-aspects-chain-ctx.nix
sini: this is related to verifying cross-scope pipes

PR #593, opened 2026-06-05
docs: update docs from source

vic:


docs/src/content/docs/releases.mdx
```diff
@@ -24,9 +24,9 @@ Den is a standalone library and has no dependencies.
 Den's `flakeModule` expects to be imported in a Nix module providing nixpkgs lib.
 
 It is very common for Den projects to use 
-[`import-tree`](https://github.com/denful/import-tree) to load their `modules/` directory.
+[`import-tree`](https://github.com/vic/import-tree) to load their `modules/` directory.
```
drupol, line 27: I guess this was not supposed to happen ?

vic: Yeah fixed in another commit. see latest on main.

vic: I believe this change was due to the llm using source code to verify docs. We might need to update repos at templates.

PR #600, opened 2026-06-08
fix: guarded empty-path route skips cleanly; honor intoPath alias

sini:
## Summary

Fixes two related bugs in simple `policy.route`, surfaced by a guarded route using `intoPath`.

### Bug 1 — guarded empty-path route crashes when the guard is false

A guarded route whose source is the raw collector module (`{ key; _file; imports }`, the empty-path case) crashed with:

```
error: The option `_file' does not exist. Definition values:
  - In `…/flake.nix':
      { _type = "if"; condition = false; content = "foo@igloo"; }
```

Two causes in `nix/lib/aspects/fx/route/wrap.nix` `guardModule`:

1. It merged the structural source module under `config`, so module metadata (`_file`/`key`/`imports`) was mis-read as option definitions.
2. It gated with `lib.mkIf` — but `mkIf false` still requires the target option to exist, so an empty-path route into an undeclared option fails type-checking even when skipped.

The forward path (`forward.nix` `guardFn`) already handles this correctly with `lib.optionalAttrs`. `guardModule` now mirrors it: gates with `optionalAttrs` (dropping the subtree entirely when false) and recurses into structural imports, leaving module metadata at module level.

### Bug 2 — `policy.route` silently drops `intoPath`

`route` only read `route.path`; an `intoPath` key (the name the forward API uses, pairing with `intoClass`/`fromClass`) was silently dropped, landing content at the class root instead of nesting. Confirmed: `intoPath = [ "wrapper" ]` placed content at the nixos root → `error: The option 'items' does not exist`.

`route` now accepts `intoPath` as the public alias for `path` (normalizing to the internal `path`), and throws `den: policy.route: pass either 'intoPath' or 'path', not both` if both are given. `path` still works.

## Tests

New regression suite `templates/ci/modules/deadbugs/route-intopath-conditional-leak.nix`:

- `test-guarded-route-skips-cleanly` — guard false → route contributes nothing, no crash.
- `test-intopath-nests` — `intoPath` nests content at the target path.

Full CI: **866/866** passing. Formatted with `just fmt`.

### DIFF
--- a/nix/lib/aspects/fx/route/wrap.nix
+++ b/nix/lib/aspects/fx/route/wrap.nix
@@ -101,18 +101,36 @@ let
       nestPlain path mod;
 
   # Wrap a module with a conditional guard.
+  #
+  # A bool guard gates content with `lib.optionalAttrs`, not `lib.mkIf`, to
+  # match the forward path (forward.nix guardFn): a false guard must contribute
+  # *nothing* — `mkIf false` still requires the target option to exist, so an
+  # empty-path route into an undeclared option would fail option type-checking
+  # even when skipped. `optionalAttrs false` drops the subtree entirely.
+  #
+  # A structural module (carries `imports`/`_file`/`key` module metadata but no
+  # flat `config`) cannot be merged under `config` — its module-level keys would
+  # be mis-read as option definitions and fail (e.g. "the option `_file' does
+  # not exist"). This is the empty-path case, where the source is the raw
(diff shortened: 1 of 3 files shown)


fmway (approved):
> Bug 2 — policy.route silently drops intoPath

I was wondering why the error in the repro is different from the one in my configuration, then I realized it's because I write `intoPath` instead of `path` 😅

Just tested in my configuration, and it works well. I think this can be merged.

PR #606, opened 2026-06-11
fix: missing host class in batteries

KP64:
With PR #605 standalone homes have a host identity. This caused the check to pass even though the host has no class causing the unfree battery to fail with the following error:
```
       error: attribute 'class' missing
       at /nix/store/wnv7bc1psj89ylp29z3kabkbpwja0npn-source/modules/aspects/batteries/unfree/unfree.nix:44:28:
           43|           && host != null
           44|           && builtins.elem host.class validClasses
             |                            ^
           45|         ) { ${host.class}.unfree.packages = allowed-names; };
```
This PR changes a host existence check to the availability of its class.

### DIFF
--- a/modules/aspects/batteries/insecure/insecure.nix
+++ b/modules/aspects/batteries/insecure/insecure.nix
@@ -38,7 +38,7 @@ let
         # nixpkgs.config.permittedInsecurePackages covers these packages.
         hostModule = lib.optionalAttrs (
           (class == "homeManager" || !builtins.elem class validClasses)
-          && host != null
+          && host ? class
           && builtins.elem host.class validClasses
         ) { ${host.class}.permittedInsecurePackages.packages = allowed-names; };
       in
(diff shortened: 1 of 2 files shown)


drupol:
I guess we need to do the same for the `insecure` battery: https://github.com/KP64/den/blob/71b764c5a9b1444586126d81bd0015315e962402/modules/aspects/batteries/insecure/insecure.nix#L41 ?

KP64:
> I guess we need to do the same for the `insecure` battery: [KP64/den@`71b764c`/modules/aspects/batteries/insecure/insecure.nix#L41](https://github.com/KP64/den/blob/71b764c5a9b1444586126d81bd0015315e962402/modules/aspects/batteries/insecure/insecure.nix#L41) ?

Correct. Now fixed :)

sini (approved):
LGTM

sini:
@KP64 thanks for the catch, by the way -- i should really try to dogfood at least one standalone home. :)

PR #623, opened 2026-06-26
feat: pipe.broadcast cross-scope push + collect reads exposed

sini:
## Summary

Adds a cross-scope **broadcast** primitive to the pipe/quirk system, makes `collect` compose with `expose`, and fixes config-thunk resolution to honour the **producing** class + scope — then makes that resolution **class-neutral** so it no longer assumes NixOS.

### `pipe.broadcast pred` — push primitive (dual of `pipe.expose`)

Where `pipe.expose` routes a value **up to the parent** applying source-side stages, `pipe.broadcast` routes the (transformed) value **laterally to every other scope matching a receiver predicate**, fleet-wide. Receivers read the pipe normally — no `collect` policy required.

```nix
den.policies.broadcast-dev = { host, user, ... }:
  [ (pipe.from "peer-dev" [
       (pipe.transform redact)                  # source-side, applied before send
       (pipe.broadcast ({ user, ... }: true))   # → every user scope, fleet-wide
     ]) ];
```

- **Receiver-only predicate** (same signature as `collectAll`); selects receivers by their own context — kind, name, tags, `host`. Compound predicates like `{ host, user, ... }: host.name == "edge"` target user scopes on a specific host.
- **Self-excluded** (`S≠R`) — a source represents itself via its own base.
- Implemented as `collectAllBroadcast` (Pass 1b) mirroring `collectAllExposed`; reuses `findMatchingAll` entity-kind filtering. `bindsPipeLocally` gains a broadcast clause so a pure receiver doesn't fall through to ancestor inheritance.

### `collect`/`collectAll` read raw **+ exposed**

`collectTagged` now reads `resolveThunks(raw) ++ allExposed.${sid}`, so a peer's collect sees data that children `pipe.expose`d up into a host. Fixes the expose-then-fleet-collect path. `allExposed` is Pass-1 data → no eval cycle.

### Producer-class config-thunk resolution

A pipe config-thunk now resolves against the **producing** class module + scope, not the consuming one (the prior behaviour was a latent bug — a host-produced thunk consumed in a home resolved against the home config).

- **Cross-scope** (broadcast/collect): the source's thunk resolves against the producer's class config.
- **Deferred** (`__configThunk`): the marker carries the producer's class + name; the class-module wrapper resolves against the producing class's config and hands each thunk both `config` (producer class) and the **owner** config (the enclosing host), mirroring home-manager. The owner arg is requested from the module system only when a marker needs it, so standalone homes keep working.
- `isConfigDependent` also detects owner-config-only thunks.

### Class-neutral resolution (no host/osConfig hard-coding)

The host↔producer navigation is **registry-driven**, not baked into core fx — so den can describe a config-owner that isn't a NixOS host (e.g. terranix, nixidy) without core edits:

- `den.classes.<class>.parentPath` — `name → path` locating a member within the enclosing config-**owner** (the same route its content is delivered to); `den.classes.<class>.parentArg` — the module arg a nested member reaches the owner by. Both `null` for **root** classes that own a top-level config (nixos, darwin, …). The home-manager battery registers `parentPath` + `parentArg = "osConfig"` from the same value as its forward-delivery path.
- `producerConfigs`/`resolveEntry` (assemble-pipes) and the class-module wrapper navigate via the registry, keyed by `scopeEntityClass`. `isConfigDependent` detects `config` or **any** registered `parentArg`. **Zero `host`/`osConfig`/`home-manager` literals remain in core fx.**
- `hostConfigs` was already generic over instantiates (any entity with an `intoAttr` output), so the model already supported non-host owners; this removes the remaining NixOS-flavored literals. (Building actual terranix/nixidy entity/class defs is follow-on work this unblocks; the two soft `or "nixos"` class-default fallbacks are left intact.)

## Tests

- **`pipe-broadcast`** (10): all-to-all · user→remote-host · source-transform · predicate scoping (incl. `{ host, user }` compound) · self-exclusion · targeted no-leak · pure-receiver binding · config-thunk from host / user source.
- **`pipe-broadcast-isolation`** (4): cross-pipe-name · host-target-excludes-home (entity-kind via shared `host` context) · no-match predicate · broadcast↔collect boundary (a fleet `collectAll` counts each raw emit once, not the broadcast-amplified view).
- **`pipe-config-scope`** (3): host→home, user→own-home, user→host producer-class resolution.
- **`pipe-scope`**: the `test-expose-then-fleet-collect` witness now passes; the `exposed-config-thunk` witness reads the owner arg.

Full CI: **1042/1042**. `treefmt` clean.

## Commits
1. `pipe.broadcast` cross-scope push + collect reads exposed
2. producer-class config-thunk resolution + broadcast review fixes
3. derive config-thunk host navigation from the class registry
4. make config-thunk resolution class-neutral (no host/osConfig literals)

### DIFF
--- a/modules/aspects/batteries/home-manager.nix
+++ b/modules/aspects/batteries/home-manager.nix
@@ -31,5 +31,7 @@ in
   den.schema.user.includes = [ result.userDetect ];
 
   den.classes.homeManager.description = "Home Manager user environment";
-  den.classes.homeManager.hostPath = userHostPath;
+  # home-manager nests under its host; a member reaches the host config via osConfig.
+  den.classes.homeManager.parentPath = userHostPath;
+  den.classes.homeManager.parentArg = "osConfig";
 }
--- a/modules/aspects/batteries/home-manager.nix
+++ b/modules/aspects/batteries/home-manager.nix
@@ -6,18 +6,20 @@
   ...
 }:
 let
+  # Where a home-manager user's config nests inside the enclosing host config.
+  # Single source of truth for both the forward delivery target and the
+  # den.classes.homeManager.hostPath the pipe layer resolves producers against.
+  userHostPath = userName: [
+    "home-manager"
+    "users"
+    userName
+  ];
   result = den.lib.home-env.makeHomeEnv {
     className = "homeManager";
     ctxName = "hm";
     optionPath = "home-manager";
     getModule = { host, ... }: inputs.home-manager."${host.class}Modules".home-manager;
-    forwardPathFn =
-      { user, ... }:
-      [
-        "home-manager"
-        "users"
-        user.userName
-      ];
+    forwardPathFn = { user, ... }: userHostPath user.userName;
     schemaIncludes = config.den.schema.hm-host.includes or [ ];
   };
 
@@ -29,4 +31,5 @@ in
   den.schema.user.includes = [ result.userDetect ];
 
   den.classes.homeManager.description = "Home Manager user environment";
+  den.classes.homeManager.hostPath = userHostPath;
 }
--- a/modules/options.nix
+++ b/modules/options.nix
@@ -32,6 +32,17 @@ let
         type = lib.types.nullOr lib.types.raw;
         default = null;
       };
+      options.hostPath = lib.mkOption {
+        description = ''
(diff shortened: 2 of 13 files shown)


drupol (approved):
Not on a computer at the moment, can't read the code efficiently. That said, the feature looks nice!

PR #625, opened 2026-06-27
fix: surface spawn-projected quirk emits at the requesting scope

sini:
## Problem

A deferred `spawn` policy (e.g. the `host-aspects` battery) projects an aspect into a post-walk nested scope. The aspect's static quirk emit only materialized inside the spawned node, so a pipe policy (`broadcast`/`collect`/`expose`/local) at the scope that **requested** the projection read `[]` — the projection did not behave "as if the requesting scope included the aspect directly".

(Surfaced by replicating `~/.claude` across a fleet via Syncthing: a user-scope `pipe.broadcast` of a host-aspects-projected `replicateHome` quirk returned empty.)

## Scope

The PR began as spawn-projected quirk surfacing and grew to cover the whole path required to make cross-host quirk delivery work end to end:

1. **Surface spawn-projected quirk emits** at the requesting scope.
2. **Resolve parametric pipe values** that need system args (e.g. `pkgs`) — deferred as config-thunks, resolved natively when handed to the module consumer.
3. **Broaden `isConfigDependent`** so any thunk arg outside `{ lib } ∪ scopeCtx` defers as a `__configThunk` instead of being force-resolved.
4. **Restructure the drain wiring** cycle-free (single shared `homeNodeSpawns`; `importsForPipes` seed; the B′ hostConfigs-null pass).
5. **Dedup the B′ per-host edges** against the host-own pass in the production delivery-edge trace.

## Fix

**Quirk surfacing.** The projected quirk emit is *static* (no fleet dependency), so it is surfaced before assembly:
- `spawn-node.nix`: the spawn return gains `quirkEmits` — the spawn subtree's non-host-bound quirk keys (host-bound ones are stripped and inherit the host's value).
- `resolve.nix`: hoist spawn materialization into one shared `homeNodeSpawns` (collapsing a duplicated `spawnNode` sub-walk across the two `mkDrained` passes), then layer its `quirkEmits` over the raw imports into a **separate** `importsForPipes` map fed to both `assemblePipes` passes. `parentState` keeps the unmodified raw map, so the spawn's own internal assembly never re-reads the surfaced quirk (cycle + internal double-count avoidance). The spawn root is absent from the pre-drain scope universe, so the quirk lands **exactly once** at the requesting scope. Multi-class spawns surface the emit once (first class that carries it wins).

**Parametric pipe config-thunks.** `assemble-pipes.nix` / `class-module.nix`: a pipe value that reads only pipeline-parametric args (host/user/…) resolves eagerly from scope context; one reading `config`/`pkgs`/parent-config args is marked `__configThunk` and resolved when the consuming module is evaluated (so a `build-info` pipe needing `pkgs` resolves to a value in the consumer, not a raw function passthrough).

**Drain restructure.** `augmentedScopeContexts` is built cycle-free from the pre-drained import map; a hostConfigs-null pass (`augmentedScopeContextsNoCfg` / `drainedForHostConfigs`) drains B′ cross-host peer configs without the assembly → augmented → hostConfigs → drain cycle.

**B′ edge dedup.** The production delivery-edge union appends `bprimeEdges` after `perHostEdges`, but `sortEdges` only sorts. `bprimeEdges` re-runs the full per-host projection, so on overlap it re-emitted identical host-scope folds (doubling them once `isConfigDependent` broadening made `hostConfigs` non-null for pipe fixtures). Now filtered by `edgeSortKey` (`target|path|source|mode`) against `perHostEdges`, keeping only the B′ cross-host delta. Bag-safe: legitimate intra-`perHostEdges` duplicates (e.g. an `os → nixos` fold pair) are untouched.

## Tests

Full suite green: **1052/1052**.

- `pipe-projection` (7) — collect/expose/local/broadcast of a projected quirk once; host-include + projection counted once each; host-bound quirk NOT surfaced; multi-class spawn surfaces once; per-user parametric projection.
- `pipe-broadcast` (11) — basic/self-excluded/targeted/predicate/source-transform/pure-receiver/to-remote-host + config-thunk (host & user).
- `pipes` — parametric flip: a `pkgs`-needing pipe value now resolves in the consumer (was raw passthrough).
- `delivery-edges.test-topology-fleet-pipe` — asserts the exact trace `count = 9`; guards the B′ double-fold (13 → 9) **and** its bag-safety (the `os` fold pair must survive → 9, not 8).
- `bprime-basedrain-crosshost` — the cross-host peer-config drain.

## Review notes

Independent review of the full branch: no blockers; mechanism sound and well-tested. Corrections applied in `4c4eaef`:
- removed dead `parentArgNames`/`readsParentArg` (superseded by the rewritten `isConfigDependent`);
- corrected the `hasAnyConfigThunk` fallback comment (a conservative over-trigger, perf-only) and the realigned `phase1` header comment.

**Follow-up (not blocking):** the cross-host `config.identity` deferral gate in `class-module.nix` is currently INERT — no consumer wires `config.identity`, so the eager branch always runs. It is a forward hook for the cross-host case; wiring it (or cutting it) plus a covering fixture is deferred.

### DIFF
--- a/nix/lib/aspects/fx/assemble-pipes.nix
+++ b/nix/lib/aspects/fx/assemble-pipes.nix
@@ -44,20 +44,12 @@ let
       let
         a = builtins.functionArgs val;
       in
-      a ? config || readsParentArg a
+      a ? config || readsParentArg a || a ? pkgs || a ? inputs' || a ? osConfig
     );
 
   # Pipeline-parametric values require pipeline context args (host, user, etc.)
   # but neither config nor a parent-config arg. Resolved eagerly via scope context.
-  isPipelineParametric =
-    val:
-    builtins.isFunction val
-    && (
-      let
-        a = builtins.functionArgs val;
-      in
-      !(a ? config) && !(readsParentArg a)
-    );
+  isPipelineParametric = val: builtins.isFunction val && !isConfigDependent val;
 
   # Resolve a local pipeline-parametric value eagerly using scope context.
   # These are quirk values like `{ host, ... }: { addr = host.addr; }` that
@@ -867,14 +859,32 @@ let
       scopeEntityClass ? { },
       hostConfigs ? null,
     }:
+    let
+      # Inherit parent scope contexts recursively
+      enrichedScopeContexts = lib.genAttrs (builtins.attrNames scopeContexts) (
+        scopeId:
+        let
+          ownCtx = scopeContexts.${scopeId} or { };
+          pid = scopeParent.${scopeId} or null;
+        in
+        if pid == null || pid == scopeId then
+          ownCtx
+        else
+          let
+            parentCtx = enrichedScopeContexts.${pid} or { };
+            inherited = lib.filterAttrs (k: _: !(ownCtx ? ${k})) parentCtx;
+          in
+          ownCtx // inherited
+      );
+    in
     if pipeNames == [ ] then
-      scopeContexts
+      enrichedScopeContexts
     else
       let
         # Pass 1: Collect all exposed data bottom-up.
         allExposed = collectAllExposed {
+          scopeContexts = enrichedScopeContexts;
           inherit
-            scopeContexts
             scopedClassImports
             scopedPipeEffects
             scopeParent
@@ -884,8 +894,8 @@ let
 
         # Pass 1b: Distribute broadcast data laterally (push, fleet-wide).
         allBroadcast = collectAllBroadcast {
+          scopeContexts = enrichedScopeContexts;
           inherit
-            scopeContexts
             scopedClassImports
             scopedPipeEffects
             scopeParent
@@ -940,7 +950,7 @@ let
                 resolvedBase = builtins.concatMap (resolveLocalParametric scopeCtx) baseValues;
                 # Own emits are produced at THIS scope; exposed values keep the
                 # producer tag set at their exposing node (re-mark is a no-op).
-                producer = producerOf scopeEntityClass scopeContexts scopeId;
+                producer = producerOf scopeEntityClass enrichedScopeContexts scopeId;
                 markedBase = markConfigThunks producer resolvedBase;
                 exposedValues = exposedForScope.${pn} or [ ];
                 markedExposed = markConfigThunks producer exposedValues;
@@ -1135,7 +1145,7 @@ let
           // pipeData
           // lib.optionalAttrs hasTargeted { __pipeTargeted = pipeTargeted; }
           // lib.optionalAttrs hasConfigThunks { __pipeConfigThunks = pipeConfigThunks; }
-        ) scopeContexts;
+        ) enrichedScopeContexts;
       in
       assembled;
 in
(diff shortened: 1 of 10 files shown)


nix/lib/aspects/fx/assemble-pipes.nix
```diff
@@ -34,30 +34,23 @@ let
...
-    val:
+    scopeCtx: val:
     builtins.isFunction val
     && (
       let
         a = builtins.functionArgs val;
+        allowedKeys = [ "lib" ] ++ builtins.attrNames scopeCtx;
       in
-      a ? config || readsParentArg a
+      builtins.any (k: !(builtins.elem k allowedKeys)) (builtins.attrNames a)
```
fmway, line 48:
```suggestion
      removeAttrs a allowedKeys != {}
```
micro-optimization, you can avoid this :)

nix/lib/aspects/fx/assemble-pipes.nix
```diff
@@ -867,14 +863,32 @@ let
...
+        in
+        if pid == null || pid == scopeId then
+          ownCtx
+        else
+          let
+            parentCtx = enrichedScopeContexts.${pid} or { };
+            inherited = lib.filterAttrs (k: _: !(ownCtx ? ${k})) parentCtx;
+          in
+          ownCtx // inherited
+      );
```
fmway, line 882: How about use mapAttrs instead?
```suggestion
      enrichedScopeContexts = builtins.mapAttrs (
        scopeId: _:
        let
          ownCtx = scopeContexts.${scopeId} or { };
          pid = scopeParent.${scopeId} or null;
        in
        if pid == null || pid == scopeId then
          ownCtx
        else
          let
            parentCtx = enrichedScopeContexts.${pid} or { };
            inherited = lib.filterAttrs (k: _: !(ownCtx ? ${k})) parentCtx;
          in
          ownCtx // inherited
      ) scopeContexts;
```

nix/lib/aspects/fx/resolve.nix
```diff
@@ -387,10 +387,18 @@ let
...
+      # stays null and assemblePipes skips cross-host instantiation entirely.
+      isConfigDependent =
+        scopeCtx: val:
+        builtins.isFunction val
+        && (
+          let
+            a = builtins.functionArgs val;
+            allowedKeys = [ "lib" ] ++ builtins.attrNames scopeCtx;
+          in
+          builtins.any (k: !(builtins.elem k allowedKeys)) (builtins.attrNames a)
```
fmway, line 400: same as before
```suggestion
          removeAttrs a allowedKeys != {}
```

fmway (dismissed):
Everything looks good. I've a few minor suggestions.

nix/lib/aspects/fx/assemble-pipes.nix
fmway: wait, seems like i lost something

fmway: My bad. It should be suggestions for line 866-882.

fmway (approved):

PR #627, opened 2026-07-02
fix: preserve source entity binding in forward fallback

ZetabS:
## Problem

Refs #620.

The documented platform-specific `hmPlatforms` example can fail when a
user-scoped complex forward includes a `fromClass` whose class key never appears
in the user's aspect tree and is not explicitly registered in `den.classes`.

For example, a user may include `hmPlatforms` and define only `hmLinux` content.
The generated forward for another class, such as `hmDarwin`, then has no
collected source modules when there is no `hmDarwin` key in that user's aspect
tree and `hmDarwin` is not registered as a class.

In that case `getCollectedSource` returns `[]`, so `resolveSourceFallback`
rewalks the source aspect from the parent scope to keep fleet visibility. That
fallback used `bindings = { }`, dropping the source scope's entity binding. For
a user-scoped forward, losing the `user` binding can make the spawn collapse
into a self-parent edge:

```
den: spawnNode spawn root equals its parent scope 'host=igloo,system=x86_64-linux'
— a self-parent edge collapses policyBoundAncestor to null and yields zero fleet peers.
The seed ctx likely lost its child binding (e.g. `user`).
```

## Fix

Restore the source scope's entity binding during complex forward fallback
rewalk.

The binding reconstruction mirrors the existing
[`mkDrained` drained-spawn path](https://github.com/denful/den/blob/70dfa3a38f19a163715fbbc64d568a00f93820be/nix/lib/aspects/fx/resolve.nix#L568-L705):

https://github.com/denful/den/blob/70dfa3a38f19a163715fbbc64d568a00f93820be/nix/lib/aspects/fx/resolve.nix#L677-L679

https://github.com/denful/den/blob/70dfa3a38f19a163715fbbc64d568a00f93820be/nix/lib/aspects/fx/resolve.nix#L702-L704

The fallback now follows the same pattern: look up the source scope's entity
kind, read that entity from `scopeContexts`, and pass it as the spawn binding.

This keeps the existing parent-scope spawn behavior for fleet visibility, while
preserving bindings like `user` when the `sourceScopeId` is a child entity
scope.

## Tests

Added `deadbugs.hm-platform-forward` covering:

- the docs `hmPlatforms` shape: active `hmLinux` forwards, inactive `hmDarwin`
  stays inert
- an undeclared/unused source class forwarded into `homeManager`, which should
  be inert rather than crash

Both new regression tests fail on the [test-only commit](https://github.com/denful/den/commit/6cc413d117483a053bad3c2e9eabc8a041b77c63) with `spawnNode spawn root equals its parent scope` and pass with this fix.

Validated locally:

- `just ci deadbugs.hm-platform-forward.test-doc-example-forwards-linux-only`
- `just ci deadbugs.hm-platform-forward.test-unknown-source-class-to-home-manager-is-inert`
- `just fmt`
- `just ci`: **1036/1036 successful**

## AI assistance

I used AI assistance while preparing this contribution:

- `ChatGPT 5.5 Thinking` helped me understand the complex forward fallback path,
  analyze the dropped source scope entity binding, and notice the existing
  `mkDrained` binding reconstruction pattern.
- `Codex GPT-5.5 medium` helped with contribution workflow, test strategy,
  fail/pass comparison, test naming/comments, and English wording.

I also used AI to translate and polish English because I am not fluent in
written English. I reviewed the final text, understand the code changes and PR
description, removed repetition, kept the diff minimal and focused, and ran the
final tests locally.

I `Gyeongmin Song`, explicitly acknowledge the following:

### DIFF
--- a/nix/lib/aspects/fx/edges/spawn-context.nix
+++ b/nix/lib/aspects/fx/edges/spawn-context.nix
@@ -1,0 +1,17 @@
+{ ... }:
+{
+  # Fallback rewalk starts from the parent scope for fleet visibility, so restore
+  # the source scope's own entity binding (for example, `user`) explicitly.
+  mkSourceScopeBindings =
+    {
+      scopeContexts,
+      scopeEntityKind,
+      sourceScopeId,
+    }:
+    let
+      sourceCtx = scopeContexts.${sourceScopeId} or { };
+      ownKind = scopeEntityKind.${sourceScopeId} or null;
+      ownRecord = if ownKind == null then null else sourceCtx.${ownKind} or null;
+    in
+    if ownRecord == null then { } else { ${ownKind} = ownRecord; };
+}
--- a/nix/lib/aspects/fx/edges/materialize-unified.nix
+++ b/nix/lib/aspects/fx/edges/materialize-unified.nix
@@ -110,6 +110,7 @@ let
         scopeContexts
         scopeParent
         scopeIsolated
+        scopeEntityKind
         ;
       # Stable scope name for the trace-edge construction (ordering only).
       name = den.lib.aspects.fx.edges.edge.scopeName {
@@ -176,6 +177,7 @@ let
               rootScopeId
               scopeContexts
               scopeParent
+              scopeEntityKind
               spawnNode
               buildForwardAspect
               ;
--- a/nix/lib/aspects/fx/edges/route.nix
+++ b/nix/lib/aspects/fx/edges/route.nix
@@ -31,6 +31,7 @@ let
     synthesize
     rootTarget
     ;
+  inherit (import ./spawn-context.nix { inherit lib; }) mkSourceScopeBindings;
   inherit (import ../scope-walk.nix { inherit lib; }) subtreeScopes;
 
   # ===== materialization mechanics (ported from route/wrap.nix) ==========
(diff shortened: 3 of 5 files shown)


ZetabS:
Sorry, I forgot to add the `ai-assisted` tag.

nix/lib/aspects/fx/edges/spawn-context.nix
```diff
@@ -0,0 +1,17 @@
...
+      scopeContexts,
+      scopeEntityKind,
+      sourceScopeId,
+    }:
+    let
+      sourceCtx = scopeContexts.${sourceScopeId} or { };
+      ownKind = scopeEntityKind.${sourceScopeId} or null;
+      ownRecord = if ownKind == null then null else sourceCtx.${ownKind} or null;
+    in
+    if ownRecord == null then { } else { ${ownKind} = ownRecord; };
```
drupol, line 16: Perhaps we can use `lib.optionalAttrs` here ? I don't know if `lib` is available.

ZetabS: `lib` is available here through the `route.nix` import. I can switch this to `lib.optionalAttrs` if you prefer.

drupol: This is an opinionated choice, I leave it up to you and the maintainer. Both are technically valid, so I can't really complain. It's just that I personally prefer using it.

sini (approved):
This looks solid enough to me.

sini:
I'll merge it once CI passes. Thank you for your contribution! :)

PR #633, opened 2026-07-17
fix: surface the requesting scope's own quirk emits into host-aspects spawns

sini:
## Problem

A quirk produced within a user's class context (a `homeManager` aspect's bare quirk key) does not reach that user's `homeManager` consumer when the consumer is projected onto the user via `den.batteries.host-aspects`. The projected consumer sees only host-scope emits (re-emitted host → home), never the user's own — asymmetric with host ↔ nixos, where a host's quirk reaches the host's nixos consumer directly.

Surfaced downstream: an overlay declared on a user-scoped `homeManager` aspect never reached that user's `nixpkgs.overlays`, so `pkgs.<x>` provided by the overlay was undefined in the user's home-manager evaluation. It "worked" only because the aspects carrying such overlays happened to be host-level roles (re-emit covered them); a genuinely user-scoped overlay broke.

## Cause

The host-aspects `policy.spawn` roots the projected consumer with `from` = the requesting scope's parent = the host scope, and links the spawn root's parent to `from`. So the consumer's quirk collection walks the *host* parent chain and never traverses the requesting *user* scope — its own directly-included quirk emits are outside the walk. #625 added the **upward** half (surface the spawn's emits up into the requesting scope's imports); the **downward** half was missing.

## Fix

The downward dual of #625.

- `spawn-node.nix` takes the requesting scope's own raw imports (`requestingImports`) and merges its non-host-bound quirk emits into the spawn root's imports used for assembly (`mergedClassImports`) — **not** into `spawnedClassImports`, which feeds the upward `quirkEmits` return (folding them there would double-count the emit back at the requesting scope). Host-bound quirks are excluded (they inherit the host's policy-assembled value, matching the existing strip).
- `resolve.nix` passes `scopedClassImportsRaw.<scopeId>` through as `requestingImports`.
- `from` = host is untouched, so #625's fleet-peer collection is preserved.

## Tests

New suite `user-class-quirk-exposure`:
- `test-host-quirk-local-nixos` — host ↔ nixos baseline
- `test-user-quirk-local-home` — user ↔ own home baseline
- `test-user-quirk-into-host-aspects-projected-consumer` — the repro (failed pre-fix: consumer read `[]`)

Full suite green: **1057/1057**.

### DIFF
--- a/nix/lib/aspects/fx/resolve.nix
+++ b/nix/lib/aspects/fx/resolve.nix
@@ -621,6 +621,12 @@ let
                   bindings = {
                     ${ownKind} = ownRecord;
                   };
+                  # The requesting scope's OWN raw imports — spawnNode surfaces its
+                  # quirk emits DOWN into the projected consumer (dual of the
+                  # `quirkEmits` surfacing UP into importsForPipes below), so a
+                  # user's directly-included quirk reaches its host-aspects-projected
+                  # homeManager consumer.
+                  requestingImports = scopedClassImportsRaw.${scopeId} or { };
                 }
               );
             };
--- a/nix/lib/aspects/fx/spawn-node.nix
+++ b/nix/lib/aspects/fx/spawn-node.nix
@@ -33,6 +33,7 @@ in
       class,
       aspect,
       bindings ? { },
+      requestingImports ? { },
     }:
     let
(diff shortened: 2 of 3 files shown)


sini:
Going to go ahead and merge this fix directly as it builds on #625

PR #641, opened 2026-07-29
fix: keep same-username homes on different hosts distinct

nolith:
Hello, I'm migrating my nix config to den and found that when home-manager and the OS are managed as separate outputs, a few strange things will happen.

Two standalone home configs that share a username but target different hosts (`user@hostA`, `user@hostB`) were stepping on each other: each flake output ended up with the other's configuration.

The cause was that a home's `name` is deliberately force-set to the bare username, and den used that `name` as an entity's identity in two separate places: the `id_hash` used by the edge layer and the scope string built by `mkScopeId` in the resolution pipeline. So both homes looked identical and collapsed onto one another. 

This PR gives entities a distinct internal identity (`__scopeName`, the full registry key like `user@hostA`) and teaches both keying layers to use it, so same-username homes on different hosts stay cleanly separated while their user-facing name and description are untouched. 

It ships as two commits (one per keying layer, since both are needed) plus regression tests covering the `id_hash` and scope paths.

### DIFF
--- a/nix/lib/aspects/fx/pipeline.nix
+++ b/nix/lib/aspects/fx/pipeline.nix
@@ -107,6 +107,17 @@ let
 
   # mkScopeId: injective scope identity from a context attrset.
   # Produces a canonical comma-separated "key=value" string, sorted by key.
+  #
+  # An entity's `name` is NOT always injective across entities of one kind: a
+  # home's `name` is force-set to the bare user name
+  # (nix/lib/entities/home.nix), so two standalone homes `user@hostA` /
+  # `user@hostB` both rendered `home=user` and collapsed onto ONE scope — their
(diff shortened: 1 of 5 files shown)


vic:
at first glance, changes look good to me, however I'd like to know if @sini is ok with them, my only question is if we do have to keep the `name = mkForce` (is forcing it -as we currently do- actually needed?, would removing that mkForce avoid we introducing a new attr?

sini:
A good question, @vic. Not one I think we need to block acceptance on. Given my intention is to replace this system soon I'm inclined to simply accept this contribution as-is.

sini (approved):

vic (approved):

nolith:
Thank you very much for the quick review

PR #653, opened 2026-08-11
fix: keep every definition of a nested aspect key defined across files

sini:
## Report

An aspect declared under the same name in two files lost one file's config. From the user report:

> A couple days ago I moved my config to den. Now I've noticed that my rollback service no longer exists. The problem seems to be that I was using the same aspect name for both. When I rename the rollback script aspect from `den.aspects.libraries.preservation` to `den.aspects.libraries.rollback` it works.

One file defined `den.aspects.libraries.preservation` as a parametric function (`{ user, ... }: { nixos = ...; }`), the other as a plain attrset. The attrset side vanished silently — no error, no warning.

## Root cause

A nested aspect key collects all of its definitions in a content wrapper's `__contentValues`. `providerType.merge` reduced that wrapper to a single definition before dispatch: with exactly one parametric function among the definitions it took that function and discarded every static attrset; otherwise it passed the wrapper through, whose forwarded attrs carry only the static side, discarding every function. Either way one side was lost with no diagnostic.

`mergeMixed` already handles a fn/attrset mix correctly by coercing functions to `includes` — it was never reached, because the collapse ran first.

Scope, as tested rather than assumed: top-level aspect names are unaffected, and so is a `_`-provides sub-aspect **declared inline** (what `issue-448-mixed-merge.nix` covers). A `_` slot assigned from a nested aspect key does route through this path. The defect is not specific to the `user` argument — a `{ host, ... }` definition drops the static side identically.

## Fix

Keep the content wrapper as a single definition and move its parametric definitions into `includes` — the shape `wrapChild` already produces for an includes element, which `listOf` hands over without reaching this merge. Both sides survive, and identity injection applies unconditionally because there is exactly one def to carry it.

An earlier revision of this branch split the wrapper into one def per definition. Review found that this loses `__provider` identity whenever every definition is parametric, trading the silent drop for silent duplication — content landing once per include path. That is fixed here and pinned by tests.

## Also fixed in this branch

**Bare parametric functions in the parametric branch** (`d77091de`). That branch coerced `__fn` wrappers to `includes` but concatenated the remaining defs raw, so a bare parametric function reached `aspectSubmodule.merge` and died with `attribute 'host' missing`. Verified pre-existing: the repro fails at `6cc4153a` too, because the old collapse also emitted a bare function whenever a wrapper held exactly one. The expansion only changed the multi-function shape — silent drop before, that same error after.

**Invented `__contentValues`** — a navigated child carries `__provider` with no `__contentValues`; giving it an empty one makes it re-flatten to nothing instead of failing.

**Coercion domain** — `lib.isFunction` is true for any attrset with `__functor`, and every merged aspect has one, so sharing one coercion demoted ordinary aspect-valued defs to `includes`. The parametric branch now uses `builtins.isFunction`, which matches exactly the bare lambda that can reach the module system as a module.

## Simplification (`09693d74`)

No behaviour change: `isParametricContent`, `soleFunctor`, and `locName` each existed in two-to-four verbatim copies across `types.nix` and `normalize.nix`; `resolveInc` re-inlined `isParametricWrapper`. Two comments that no longer described the code are corrected. Both `__functor` ambiguity checks are kept deliberately — the mixed branch erases `__functor` before the submodule merge could see the conflict.

## Tests

- `samename-mixed-merge-drop.nix` — the reported shapes plus controls for static+static, distinct names, and a top-level aspect name.
- `nested-key-wrapper-identity.nix` — identity/dedup for all-parametric nested keys; two tests red before the final commit, two controls green throughout.
- `parametric-wrapper-beside-content-wrapper.nix` — the bare-function-in-parametric-branch regression.
- `quirk-set-and-consume-same-aspect.nix` — the reporter also suspected quirks, believing that setting a quirk in the aspect that consumes it was dropped. That diagnosis does not hold: those cases were already green. Their quirk symptom was this merge defect, the value sitting in the static definition that the parametric one displaced.

## Verification

`just ci` 1101/1101 · `just all check` (8 templates) · `just check-noflake` · `just unit` · `just fmt` clean.

### DIFF
--- a/nix/lib/aspects/default.nix
+++ b/nix/lib/aspects/default.nix
@@ -122,6 +122,7 @@ in
   # Predicates exported directly (not through types mapAttrs which applies { } to each value).
   inherit (rawTypes)
     isParametricWrapper
+    isParametricContent
     isSubmoduleFn
     isMeaningfulName
     isSyntheticName
--- a/nix/lib/aspects/fx/aspect/normalize.nix
+++ b/nix/lib/aspects/fx/aspect/normalize.nix
@@ -4,7 +4,12 @@
   den,
 }:
 let
-  inherit (den.lib.aspects) isSubmoduleFn isMeaningfulName isParametricWrapper;
+  inherit (den.lib.aspects)
+    isSubmoduleFn
+    isMeaningfulName
+    isParametricWrapper
+    isParametricContent
+    ;
 
   # Normalize a NixOS module function into an aspect attrset via type merge.
   normalizeModuleFn =
@@ -98,16 +103,7 @@ let
       let
         prov = child.__provider or [ ];
         provName = if prov != [ ] then lib.last prov else null;
-        fns = builtins.filter (
-          d:
-          lib.isFunction d.value
-          && (
-            let
-              args = builtins.functionArgs d.value;
-            in
-            args != { } && !(args ? config) && !(args ? options)
-          )
-        ) (child.__contentValues or [ ]);
+        fns = builtins.filter isParametricContent (child.__contentValues or [ ]);
       in
       child
       // lib.optionalAttrs (provName != null) {
--- a/nix/lib/aspects/types.nix
+++ b/nix/lib/aspects/types.nix
@@ -174,20 +174,31 @@ let
     };
 
   # A parametric function reaching aspectSubmodule.merge is evaluated as a NixOS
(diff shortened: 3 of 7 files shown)


sini:
Given this a live bug fix I went ahead and merged it. If there are any review comments I'll address them in followup.

theutz:
I pulled it down and used it in my flake without any issues!

Issue #629, opened 2026-07-07
can't request `user` inside an aspect's `nixos` class module


lucasshiva:
I was creating an aspect for Cosmic DE and realized I needed the user's name to enable autologin. I knew we could request either `host`, `user`, or both, so I requested user, but the `nixos` class module was skipped. I had this:

```nix
{
  den.aspects.desktop.cosmic = {
    nixos = { user, config, ... }: {
      # entirely block is skipped
    };
  };
}
```

I checked the [class modules](https://den.denful.dev/explanation/class-modules/) page in the docs, but I didn't see any examples  of the `nixos` class module requesting only `user`. I did, however, see examples adding `host` as well, so I tried doing the same:

```nix
{
  den.aspects.desktop.cosmic = {
    nixos = { host, user, config, ... }: {
      # still skipped
    };
  };
}
```

No success. The only way to get something similar to the above working was to request `user` at the aspect level, like shown in the [parametric](https://den.denful.dev/explanation/parametric/) page in the docs:

```nix
{
  den.aspects.desktop.cosmic = { user, ... }: {
    nixos = { config, lib, ... }: {
      # No longer skipped.
      services.displayManager.cosmic-greeter.enable = true;
      services.desktopManager.cosmic.enable = true;

      services.displayManager.autoLogin = {
        enable = true;
        user = user.userName;
      };
    };
  };
}
```

For context, I am on NixOS, and this particular aspect doesn't have a `homeManager` class module. I also have options inside the `nixos`module, so it was easy to realize the entire module was being skipped because my options didn't register. The options were also the reason why I couldn't use `provides.to-users = { user, ... }` since it's outside the `nixos` class module and thus wouldn't have access to its `config`.

I'm not sure if this is a bug or me understanding parametric aspects wrong. The docs made it seem as if `nixos = { host, user, ... }` would work as long as we were in a context where host is present, which I am. I also assume `user` is always present because it seems we can request it from both system and home class modules, and because we can't really have a system without an user, but do correct me if I'm wrong on how `Den` approaches that.

sini:
I'll see if I can reproduce this scenario as a unit test -- can you confirm if there is in fact a user resolving on the particular host you're seeing this on? (Because obviously no user then no parametric expansion explaining the behavior.) And confirm that this aspect is indeed in the aspect tree?

[...]

PR #659, opened 2026-08-25
docs: cover undocumented public surface and fix stale API references

sini:
Documents public features that shipped without docs, and fixes code comments still pointing at the removed adapter API.

## New page

**`explanation/structural-introspection`** — `entity.hasAspect` / `entity.aspects`, `meta.handleWith` with `den.lib.aspects.fx.constraints` (`exclude` / `substitute` / `filterBy` plus their `.global` variants), and the lib-level building blocks (`hasAspectIn`, `collectPathSet`, `mkEntityHasAspect`, `mkProjectedHasAspect`). Covers where `hasAspect` is safe to call, and why `includes` is not.

Previously this API was only discoverable by reading `templates/example`.

## Reference and guide additions

- `pipe.broadcast` stage semantics + a worked fleet example (had zero docs)
- Conditional aspects (`meta.guard` / `meta.aspects`) via `policy.when`
- `den.batteries.flake-scope`
- Built-in policy tables: `os-to-host`, `user-to-host`, the home-env family, wsl, and the flake-parts pair
- `den.schema` collection fields (`isEntity`, `isolated`, `parent`, `collisionPolicy`) and the `den.schema.aspect` base module
- `den.classes` fields: `description`, `forwardTo`, `parentPath`, `parentArg`
- `den.reservedKeys`, `den.systems`, flat host/home declaration forms
- Policy `deliver`/`route` `guard` + `adaptArgs`, `route.collectSubtree`, the `intoPath` alias, custom `instantiate` specs, `resolve.shared.to`
- `captureFleet`, `resolveWithPaths` / `resolveWithState`
- Class-module arg promotion for descendant entity kinds (#629)

## Corrections to existing docs

- Class-module collision policy resolves over **four** levels, not three (aspect → per-entity → `den.schema.<kind>` → global), matching `resolveCollisionPolicy` in `class-module.nix`
- Stale battery anchors in the sidebar (19 links) and in both tutorials
- `lib.mdx` linked to a nonexistent `/reference/options/` page

## Code changes

Comments and one option `description` only, no behaviour change. `modules/context/has-aspect.nix` and the example template referenced `oneOfAspects`, `excludeAspect`, `meta.adapter` and `adapters.nix` — none of which still exist — and named the filter constraint `filter`/`filterIncludes` rather than `filterBy`.

## Verification

- `nix develop -c just fmt` — clean, no changes to the edits
- `nix flake check --override-input den . ./templates/example` — `all checks passed!`
- `pnpm build` — 61 pages, clean
- Internal link/anchor check over `dist/`: **8,349 links, 0 broken pages, 0 broken anchors**

## Known follow-ups (not in this PR)

Still undocumented: `denTest`'s `expectedError` option and the `show` / `funnyNames` / `withSystem` helpers, and there is no option-reference page for `den.config`. Separately, `templates/ci/` still carries stale `test-excludeAspect-*` test names and a `meta.adapter` comment.


docs/src/content/docs/explanation/structural-introspection.mdx
```diff
@@ -0,0 +1,212 @@
...
+The two live at different points in evaluation, and that difference is what
+makes them safe.
+
+## Reading structure: `entity.hasAspect`
+
+Every entity (host, user, home — and any custom schema kind) carries a
+readOnly `hasAspect` functor computed from its resolved aspect tree:
+
+```nix
+host.hasAspect den.aspects.zfs-root        # bool — primary class
```
theutz: What is the difference between `host.hasAspect` and `host.hasAspect.forAnyClass` here? They seem identical.

docs/src/content/docs/explanation/structural-introspection.mdx
```diff
@@ -0,0 +1,212 @@
...
+```
+
+The tree depends on every aspect's `includes`; `includes` depending on the
+tree is a back-edge with no fixed point. If the decision is *structural*
+(which aspects belong in the tree), use [constraints](#writing-structure-metahandlewith)
+instead — they run during the walk, with full structural visibility, and
+operate *on* the tree rather than reading *from inside* it.
+</Aside>
+
+### What "present" means
```
theutz, line 112: The concepts of `present` and `absent` haven't been introduced yet, so this section is unclear. Is this the equivalent of the boolean returned from `<entity>.hasAspect`?

docs/src/content/docs/explanation/structural-introspection.mdx
```diff
@@ -0,0 +1,212 @@
...
+      environment.etc."aspect-inventory".text =
+        lib.concatMapStringsSep "\n" (n: n.identity) host.aspects;
+    };
+  };
+```
+
+(`aspects` lives on the *entity*, so reach it through the entity arg —
+`host.aspects`, `user.aspects`, `home.aspects` — not through `config`.)
+
+The entity's `hasAspect` record also carries `aspectsForClass` and
```
theutz: Ah. So, to me this implies that `hasAspect` only returns aspects on the entity's primary class. If that's true, that should probably be clarified [here](https://github.com/denful/den/pull/659/changes#r3855603766).

theutz: If that's not what this means, then additional clarification would be useful.

docs/src/content/docs/explanation/structural-introspection.mdx
```diff
@@ -0,0 +1,212 @@
...
+of them) and registers it for the aspect's subtree. Constraints are
+constructed via `den.lib.aspects.fx.constraints`:
+
+```nix
+den.aspects.secrets-bundle = {
+  includes = [
+    den.aspects.agenix-rekey
+    den.aspects.sops-nix
+  ];
+  # Prefer agenix: drop sops-nix from this subtree whenever both are present.
```
theutz: This API is confusing to me. You include an aspect, but then you immediately exclude it. But there's no obvious indication of why this `fx.constraints.exclude` only fires when `den.aspects.agenix-rekey` is present.

TL;DR - me confuused.

docs/src/content/docs/guides/batteries.mdx
```diff
@@ -183,6 +183,20 @@ den.default.includes = [ den.batteries.self' ];
 
 Requires flake-parts. Works in host, user, and home contexts.
 
+### `den.batteries.flake-scope`
```
theutz, line 186: Oooh. I didn't know about this one. I'm gonna use it. :D

docs/src/content/docs/guides/custom-classes.mdx
```diff
@@ -131,6 +131,43 @@ den.aspects.alice.container = {
 };
 ```
 
+## Class Schema Fields
+
+Every entry in `den.classes.<name>` is a small schema, not just a marker:
```
theutz: Is "small schema" the right language here? Seems like you're employing that a class is an entity...

docs/src/content/docs/reference/schema.mdx
```diff
@@ -108,6 +116,19 @@ den.hosts.x86_64-linux.myhost = {
 };
 ```
 
+### Flat form
```
theutz, line 119: Another good one I didn't know about!

theutz (changes requested):
A much needed refresh. This is great!

docs/src/content/docs/explanation/structural-introspection.mdx
sini: Identical in outcome — correct. Membership is *structural*, and the resolved tree doesn't vary by class (a class decides what an aspect *emits*, not whether it is *in* the tree). `templates/ci/modules/internal-api/has-aspect.nix:565` asserts exactly that: `forClass "user"` and `forClass "homeManager"` both return `true` for an aspect that only declares `homeManager` content.

So the three differ only in the class they ask at: `hasAspect` uses the primary class (head of the entity's `classes`), `forClass` uses the one you name — and is the only variant that can answer `false`, for a class the entity doesn't declare — and `forAnyClass` unions over the declared ones.

Added a paragraph after the block stating that outright, rather than implying a distinction the return values don't have.

docs/src/content/docs/explanation/structural-introspection.mdx
sini: Yes — *present* is exactly the boolean from `<entity>.hasAspect`. The section now opens by saying so, before the two rules that decide it, so *absent* isn't introduced cold.

docs/src/content/docs/explanation/structural-introspection.mdx
sini: Half of it. `hasAspect` does ask at the primary class, but it isn't limited to it: the structural tree is the same whichever class it resolves at, so which class it asks at changes nothing about the answer. Same for `entity.aspects` — taken at the primary class, but no class carries a different list.

Clarified in both places, including the line-34 block you linked.

docs/src/content/docs/explanation/structural-introspection.mdx
sini: The example was wrong, not you. `constraints.exclude` is unconditional — `nix/lib/aspects/fx/constraints.nix:12` builds a plain `{ type = "exclude"; identity = …; }` record with no presence test anywhere — so the comment claiming it drops sops-nix "whenever both are present" described behaviour that does not exist, and including sops-nix one line above only to exclude it made the whole thing read as a no-op.

Rewritten: the bundle includes a role aspect that pulls sops-nix in transitively, which is the case constraints are actually for. `secrets-bundle` can't see what that role includes, but it can declare sops-nix unwelcome anywhere below it. Added a paragraph stating the unconditionality explicitly.

docs/src/content/docs/guides/batteries.mdx
sini: Existing behaviour, just never documented — which is what this PR is for. No change here.

docs/src/content/docs/guides/custom-classes.mdx
sini: Agreed on the collision. `schema` already means `den.schema.<kind>`, the entity-kind mechanism, so "small schema" does read as "a class is an entity kind". Changed to "a record of options", and renamed the heading `Class Schema Fields` → `Class Fields` (nothing links to that anchor).

docs/src/content/docs/reference/schema.mdx
sini: Same as the batteries one — existing, previously undocumented. No change here.

docs/src/content/docs/explanation/structural-introspection.mdx
theutz: Noice. Thank you!

docs/src/content/docs/explanation/structural-introspection.mdx
theutz: Ok, good. Yeah, I hope that flow helps new readers, too. I think the concept was clear by the _end_ of the doc, but it wasn't initially.

docs/src/content/docs/guides/custom-classes.mdx
theutz: Much better.

docs/src/content/docs/explanation/structural-introspection.mdx
theutz: Oh, that's brilliant. That makes significantly more sense. :D

theutz (approved):
Love it.

Issue #663, opened 2026-08-28
`den.hosts.<system>.<name>.includes` is accepted and silently dropped


sini:
## Summary

Writing `includes = [ … ]` directly on a host entity does nothing. The key is
accepted by the freeform host submodule, never read by the resolver, and
discarded with no warning and no error. The host builds successfully with an
empty aspect tree.

This is a silent drop on a plausible, discoverable spelling — `includes` is the
activation key everywhere else in den (`den.aspects.<n>.includes`,
`den.schema.host.includes`, `den.default.includes`), so writing it on the host
entity is the natural guess.

Found while supporting #660 / discussion #662, where a third-party integration
guide used the host-level spelling in two of its three patterns. The user's
hosts evaluated cleanly and contained nothing.

## Reproduction

```nix
{ den, inputs, ... }:
{
  imports = [ inputs.den.flakeModule ];

  den.aspects.base = {
    nixos = {
      networking.hostName = "from-base";
      fileSystems."/" = { device = "/dev/null"; fsType = "auto"; };
      boot.loader.grub.enable = false;
      system.stateVersion = "25.05";
    };
  };

  # negative: per-host includes
  den.hosts.x86_64-linux.igloo = {
    includes = [ den.aspects.base ];
  };

  # positive control: aspect-level includes, same aspect, same run
  den.aspects.iceberg.includes = [ den.aspects.base ];
  den.hosts.x86_64-linux.iceberg = { };
}
```

```console
$ nix eval --impure --expr 'let f = builtins.getFlake "path:."; in {
    perHostIncludes = f.nixosConfigurations.igloo.config.networking.hostName;
    aspectIncludes  = f.nixosConfigurations.iceberg.config.networking.hostName;
  }'
{ aspectIncludes = "from-base"; perHostIncludes = "nixos"; }
```

`"nixos"` is nixpkgs' default — nothing from `den.aspects.base` reached `igloo`.

Measured against `c7ef3f1`, nixpkgs `nixos-unstable` (26.11pre1062397).

## Cause

`resolveEntity` builds the entity root aspect's `includes` from exactly two
sources — the entity's self-provide (`den.aspects.<name>` via `lookupAspect`)
and the schema-level collection:

`nix/lib/resolve-entity.nix:73`
```nix
includes = selfProvide ++ schemaIncludes;
```

Nothing reads `includes` off the entity instance. The instance type is
constructed with `strict = false`:

`nix/lib/entities/host.nix:55-57`
```nix
[...]

PR #678, opened 2026-09-08
fix: resolve identity by definition value and split the provenance field

sini:
https://github.com/denful/den/discussions/670

## Summary

Identity:

- Resolves aspect identity by comparing raw definition values at `providerType.merge`, so two owners writing their own inline aspect of one name stay distinct while a shared `let`-bound value included twice keeps one identity.
- Compares the whole raw value with no projection, because a projected comparison merges two values that differ only in the stripped part.
- Keeps a list of claims per definition position rather than a single claim, as one slot lets a differing sibling hold a position permanently and makes the outcome depend on walk order.
- Carries `__walkStamped` through the parametric-resolve round-trip, so a walk-stamped child does not have its position encoded a second time.

Policy:

- Buckets policy identity claims by scope and bare name, so two `mkPolicy "tools"` records in different aspects both fire while a shared `den.policies.foo` included from two aspects fires once.
- Resolves `excludes` by raw claim value rather than by bare name, as a displaced claimant was otherwise unaddressable and an exclude suppressed a different policy than the one named.

Alias:

- Routes the `_`-to-`provides` fold through one shared helper instead of constructing it separately per site, as `mergeFunctions`' battery branch was dropping a written `fn._` on the floor rather than folding it.
- Leaves root's `mkAliasOptionModule` out of that helper deliberately and records why, because the helper is a `//` overwrite and wiring it into root would replace root's conflict detection with spelling-priority.
- Pins that root and nested deliberately differ on a conflicting definition, so the divergence is asserted rather than discovered.

Provenance:

- Splits `providerPrefix` into `origin` for the container seed and `chain` for the threaded definition path, because one field carried two facts that coincide at a container root and so conflated without any test showing it.

Batteries:

- States each shipped predicate battery's own chain, so the `insecure` and `unfree` batteries keep their identities on a stock configuration.

Structural keys:

- Derives the `__`-prefixed half of the structural-key registry by rule instead of listing twelve markers by hand, so a new marker is structural the moment it exists rather than when someone remembers to register it.
- Replaces `normalizeRoot`'s functor whitelist with the same structural carry-forward `mkParametricBase` uses, as that branch was dropping a functor-shaped root's `excludes` outright.
- Makes the key classifier and `aspectKeyType.merge` share one predicate, because they previously disagreed about an unlisted `__` key, one treating it as a nested key and the other wrapping it as content.

Delivery:

- Records `bind`'s misplaced-entity-arg verdict as pipeline state so the post-assembly drain's residue guard covers it, since that verdict is silent and correct everywhere except a terminal position where no later scope can deliver.
- Leaves the double-cover verdict unrecorded and says why in the code, as the descendant does receive that content and no test can catch the mistake of recording it.

Entity instances:

- Honours `includes` and `excludes` written on an entity instance (`den.hosts.<system>.<name>.includes`), which were accepted and dropped in silence (#663). `resolveEntity` built the root aspect's collections from the self-provide and the schema tier only, while the instance type is built with `strict = false`, so its freeform type absorbed the key instead of raising "option does not exist" — the host built with an empty aspect tree and the only signal was a warning about a different fact. Read in `resolveEntity` rather than per entity kind, so every kind declaring `isEntity` gains it at once, and read last in the include order, an instance being the most specific site content can be written at.

Excludes:

- Flattens an aspect's `excludes` so a list-wrapped policy reference excludes, because `providerType` names a list of policy records as valid while the reader had no list arm and reduced it to `<anon>`.
- Warns when a raw-ref exclude matches nothing anywhere in the run, and only then, so a genuine typo is reported while an exclude that legitimately matches in one scope and not another stays quiet.
- Gives the two schema-tier collections ONE element check, so a list-wrapped policy reference is admitted at both tiers. `den.schema.<kind>.excludes` rejected the shape the aspect tier had just been fixed to honour, and `providerType` names a list of policy records as valid, so the divergence had no basis. The `includes` check already recursed into nested lists and admitted a function while `excludes` did neither — two copies diverging by inspection, now one check parameterised by collection name.

Provides:

- Reserves structural names from the provides forward at both raw construction sites, so a provides child named after an aspect option behaves the same at all three sites instead of landing in the aspect's own option position at two of them.

Performance:

- Resolves rawRef excludes once per dispatch rather than once per candidate policy name, as neither the registry flatten nor the claim resolution depends on the name being tested.

Tooling:

- Surfaces the evaluator's stderr in `ci.bash` and reports a dead evaluator separately from a test failure, because `total` is computed as `pass + fail` over whatever reached the JSON stream, so a run that stopped early read as a clean `N/N` with the diagnostic discarded.
- Corrects nine stale claims left by this branch and its predecessors, including a public reference documenting a field that `metaType`'s freeform now absorbs silently and a live test citing a coverage suite that no longer exists.
- Corrects three more found by reviewing the branch against its own docs: the aspect reference's structural-key table still typed `excludes` as `listOf unspecified` and still listed `meta.provider`, contradicting the caution block sixty lines below it that documents the rename; and `.claude/skills/den-debugging.md` held the last `__provider` reference in the repo, in the skill CLAUDE.md points debugging work at.

## Behaviour changes

Configurations that were silently losing content now deliver it, so a rebuild can emit modules it did not emit before:

- Two owners writing their own inline aspect under one name stay distinct instead of collapsing onto one identity, which is the defect this branch exists to close.
- Two `mkPolicy` records sharing a bare name in different aspects both fire, where one previously displaced the other.
- A functor-shaped root aspect keeps its `excludes` and other structural siblings, which `normalizeRoot` was dropping outright.
- A parametric wrapper keeps its sibling `includes` across the resolve round-trip.
- A provides child whose name collides with an aspect option survives, and is reachable through `._` at all three construction sites rather than two of them landing it in the aspect's own option position.

Input that was previously accepted and ignored is now rejected or honoured, which can change an existing configuration:

- `includes` and `excludes` on an entity instance are honoured. A configuration that wrote them and silently got nothing will start delivering that content, which is the defect #663 reports; a configuration that wrote them redundantly beside a same-named aspect now applies them twice over, deduplicated at the gate.
- `excludes = [ "some-name" ]` is a type error. It previously reduced to `<anon>`, matched nothing, and excluded nothing in silence.
- `excludes = [ [ policy ] ]` now excludes. The same value was admitted by the type and then discarded, so a configuration that relied on the no-op will start suppressing the policy it names.
- An exclude resolves against the raw claim value rather than a bare name, so it suppresses the policy actually named instead of a different one that happens to share its name.
- An aliased aspect's authored `name` outranks its nested position when the two disagree.
- Any `__`-prefixed key is treated as a pipeline internal by rule rather than by appearing on a list of twelve. A `__`-prefixed key an aspect defines is no longer dispatched as class or nested content.

New diagnostics, where the previous behaviour was silence:

- A pipe-arg-deferred include whose registered effects cannot be delivered throws, naming the aspect and the scope. It previously delivered the direct half and dropped the policy half with nothing said.
- An exclude naming a policy that never registered anywhere in a resolution emits a `den:` warning. It does not change what is excluded. A fleet-level exclude warns once per host that does not include the policy, since a resolution is one entity.
- `ci.bash` reports a dead evaluator as `💥 EVALUATOR FAILED` with the evaluator's stderr, rather than discarding it and printing a tally over whatever the run reached.

Renamed:

- `meta.provider` is now `meta.aspect-chain`, and an absent chain is distinguished from a root one (`null` against `[ ]`). `metaType` is freeform, so a stale `meta.provider` is absorbed without error rather than rejected.

## Validation

- `nix develop -c just ci`: 1201/1201 at `cb5439b`, exit 0, zero failures and zero errors, summary line present and the collected count agreeing with the numerator. (1194/1194 at `ce38772`, before the four cells the #663 fix added and the three the schema-tier check added.)
- `nix develop -c just ci performance`: 31/31, exit 0, including the `resolve.test-chain-100` canary and the entity and fleet cells this branch added.
- Compared each run against the expected baseline rather than against the printed total, because `ci.bash` computes `total` as `pass + fail`, which makes `N/N` an identity that holds on a truncated run.
- Verified the `ci.bash` change by forcing a real parse error: the evaluator's message with file, line and column now reaches stderr, where the identical run previously produced zero bytes and no summary.
- Falsified every new cell against the unfixed tree and read its actual red output rather than trusting the green.
- Checked the identity change in both directions, because over-splitting double-emits visibly while over-merging drops a definition silently, and no single cell discriminates both.
- Measured `origin` against gen's own source rather than by name: seeding it makes a nested chain read additively through the subtree, matching gen's `origin ++ path`.
- Reviewed the whole branch as one diff after the per-task reviews, which is what surfaced the `excludes` list-element drop and the provides-forward site disagreement.
- Reproduced #663 before fixing it: both instance-level cells red and both controls green on the unfixed tree, since the failure mode is "produces a plausible value" rather than "throws". Falsified the schema-tier cell the same way — it errors on the unfixed tree and is the only one of the six that flips.
- Ran `nix flake check` per template rather than reading `just all check`'s exit code, which aborts at `ci` on a `flake.packages` double declaration that reproduces unchanged on `origin/main`. Nine templates pass; `diagram-demo`, `fleet-demo` and `terranix-demo` fail for reasons that also reproduce on `origin/main`, and `diagram-demo`'s is the one this branch caused — see below.

## Known consequence outside this repo

The `meta.provider` → `meta.aspect-chain` rename left `den-diagram` reading a field den no longer emits. `graph.nix` read `entry.provider` with no fallback, so `diagram-demo` aborted with `attribute 'provider' missing`; three further sites read it through an `or [ ]` fallback and kept rendering with a silently empty chain, flattening the fleet and namespace views' provenance. Nothing in den's gates can catch this — den's CI has no den-diagram dependency by design, and `just all check` covers neither diagram template.

Fixed in denful/den-diagram#2 (merged, `3ba2e47`), which reads both spellings from one helper so a single den-diagram works against either side of the rename. The lock bump for `diagram-demo` and `fleet-demo` is NOT in this PR. Both templates still fail `nix flake check` afterwards, for the unrelated pre-existing reasons noted above.


theutz:
I'm a bit unclear if #678 is meant to be fixed by this, but if it is, I found some behavior that seemed unintuitive. A minimal example:

```nix
# file-1.nix
{
 den.hosts.igloo.includes = [ <some-aspect> ];
}

# file-2.nix
{
  den.hosts.igloo.includes = [ <some-other-aspect> ];
}
```

The list of includes doesn't seem to merge, nor does it warn that there are conflicting definitions. Just silent last-entry-wins, I suppose?

PR #682, opened 2026-09-18
fix: keep unnameable context values out of scope identity

sini:
## Summary

- Narrows `mkScopeId` to context values it can name, as an unnameable value falls through to the type placeholder branch and renders as `den=<set:den>`, so it carries no identity and only splits one logical scope across two ids depending on which module args happened to be bound on the path.
- Fixes a quirk reached through `den.hosts.<name>.includes` arriving empty while the same aspect through `den.aspects.<name>.includes` delivers, because pipe emits are bucketed by scope id and `assemble-pipes` gives a consumer its own bucket, with no plain ancestor inheritance of raw emits outside `policyBoundAncestor`.
- Leaves the keys in the context itself, where binding still reads them, so only the identity is narrowed.
- Accounts for class content being unaffected, since it is not bucketed by scope id. In the reporting configuration two aspects delivering `darwin.homebrew.casks` through the same instance list arrived while one delivering a `casks` quirk did not.
- Adds fifteen cells covering quirk collection through both entrypoints, including the darwin class and the flat `den.hosts.<name>` spelling that no existing cell reached.

## Validation

- `nix develop -c just ci`: 1236/1236, exit 0, read off stderr.
- Falsified the new unit cell against the unfixed tree. It returns `"host=igloo,inputs=<set:inputs>,lib=<set:lib>,user=tux"` where it expects `"host=igloo,user=tux"`, and its control cell stays green in that same run, so the cell measures the exclusion rather than a scope id that dropped every key.
- Reproduced the report end to end on the reporter's own configuration against den `20d1e76`, changing one token in one file: `den.aspects.kocaeli.includes` yields the tap, `den.hosts.kocaeli.includes` yields `[ ]`, both at exit 0 with no error.
- Re-ran that same pristine tree and that same one token change against this branch, which yields the tap.
- Traced the scope pushes on that host. The host scope pushed ctx keys `host|system` while the user scope pushed `den|host|inputs|lib|system|user`, which is where the placeholder keys entered the identity.
- Confirmed the class versus quirk contrast inside the reporter's config using his own aspects, where `slack` and `spotify` reach `homebrew.casks` through that instance list via a class key and `rootshell` does not via a quirk key.


theutz:
CleanShot 2026-09-18 at 21 54 22@2x

CleanShot 2026-09-18 at 21 54 59@2x

theutz (approved):

